PR #43, opened 2023-07-04
Fix colors of inscription view in vscode

ivy-lli:
- Remove vscode variables, use colors of inscription view (same as process editor)


ivy-cst (commented):
Cool, thanks a lot :+1:

ivy-cst (approved):

PR #47, opened 2023-07-13
Sending shutdown seems to work 

weissreto:


ivy-lmu (approved):
cool, thanks!

vscode-extensions/engine/extension/src/index.ts
```diff
@@ -25,12 +25,13 @@ async function startEmbeddedEngine(extensionUri: vscode.Uri) {
   const env = {
     env: { ...process.env, JAVA_OPTS_IVY_SYSTEM: '-Divy.enable.lsp=true -Dglsp.test.mode=true -Divy.engine.testheadless=true' }
   };
-
+  console.log('Start ' + engineLauncherScriptPath);
```
ivy-lmu, line 28: here you could also append to `outputChannel`

ivy-lmu:
> What do you think should I let the console stuff in or remove it?
I rarely make use of it, but here it makes sense I think

ivy-lmu (approved):

PR #54, opened 2023-08-04
fix: add QueryClientProvider & pass pmv and app

ivy-lmu:
image

### DIFF
--- a/vscode-extensions/inscription/webview/src/index.tsx
+++ b/vscode-extensions/inscription/webview/src/index.tsx
@@ -6,6 +6,7 @@ import ReactDOM from 'react-dom/client';
 import './index.css';
 import * as reactMonaco from 'monaco-editor/esm/vs/editor/editor.api';
 import { SelectedElement } from '@axonivy/vscode-base';
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
 
 declare var acquireVsCodeApi: any;
 var vscode = acquireVsCodeApi();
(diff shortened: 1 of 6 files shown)


yarn.lock
```diff
@@ -18,40 +18,40 @@
     "@jridgewell/gen-mapping" "^0.3.0"
     "@jridgewell/trace-mapping" "^0.3.9"
 
-"@axonivy/config-editor@~11.2.0-next.28":
```
weissreto, line 21: Do we need to have this file to be in the repository? Seems to be house keeping for yarn.

vscode-extensions/inscription/webview/src/index.tsx
```diff
@@ -78,13 +79,21 @@ function startInscriptionClient(webSocketAddress: string) {
 }
 
 export function render(inscriptionClient: InscriptionClient, selectedElement: SelectedElement = undefined): React.ReactElement {
+  const queryClient = new QueryClient();
```
weissreto, line 82: It seems to be complex to integrate the inscription view app now. We need to know stuff about the react-query. 
I would like to see such things hidden in the inscription view app itself. 
Also the configuration with the three different parameters could be simplified by providing the selectedElement in one arg directly.

weissreto (approved):
Some architectural comments that can be solved in later PRs.
Good to merge to fix the broken integration now.

yarn.lock
ivy-lmu: yes, we want to have that file in the repo to get consistent installs across machines. see https://classic.yarnpkg.com/lang/en/docs/yarn-lock/

yarn.lock has changed slightly as I increased our own dependencies like @axonivy/config-editor and added new ones. if no dependencies are touched, it does not change.

vscode-extensions/inscription/webview/src/index.tsx
ivy-lmu: good ideas

ivy-lli: I agree, that this it is not ideal that the vscode extension needs to know this stuff here! 
We definitly should move this more into the inscription view code, so this stuff here is not needed.
I wasn't aware of this, sorry.

PR #58, opened 2023-08-09
new variables.yaml command improvement

ivy-lmu:
- remove command from panel
- create new variables.yaml under /config

### DIFF
--- a/vscode-extensions/config-editor/extension/src/editor-provider.ts
+++ b/vscode-extensions/config-editor/extension/src/editor-provider.ts
@@ -1,9 +1,10 @@
 import * as vscode from 'vscode';
 import { getNonce, getUri } from './util';
+import * as fs from 'fs';
 
 export class YamlEditorProvider implements vscode.CustomTextEditorProvider {
-  private static newFileId = 1;
   private static readonly viewType = 'yaml-variables-editor';
+  private static readonly fileName = 'variables.yaml';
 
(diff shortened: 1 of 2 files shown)


vscode-extensions/config-editor/extension/src/editor-provider.ts
```diff
@@ -1,25 +1,32 @@
...
+      if (!fs.existsSync(configPath.fsPath)) {
+        vscode.window.showErrorMessage(`No config directory found in the workspace`);
+        return;
+      }
+      const variablesPath = vscode.Uri.joinPath(configPath, this.fileName);
+      if (fs.existsSync(variablesPath.fsPath)) {
+        vscode.window.showErrorMessage(`${this.fileName} file already exists`);
+        return;
+      }
+      vscode.workspace.fs.writeFile(variablesPath, new TextEncoder().encode('Variables: {}'));
```
ivy-lli, line 28: I know the yaml-js lib does write an empty map with {} right? But I would create the new file without this :)

ivy-lli (approved):
Thanks :)

vscode-extensions/config-editor/extension/src/editor-provider.ts
ivy-lmu: but you are right, `Variables:` is better ;)

PR #62, opened 2023-08-11
XIVY-12086 fix: engine activation

ivy-lmu:
otherwise inscription and process-editor trigger the engine extension to start


ivy-cst (approved):
:+1: 
Would it be not better on 'ivy:hasIvyProjects'?

ivy-lmu:
> 👍 Would it be not better on 'ivy:hasIvyProjects'?

seems not to be possible

PR #68, opened 2023-08-30
Fix outline for theia

ivy-lli:
- Send select actions direct to client and reveal the webview afterwards
  This is needed because the webview loose the focus on theia and will not receive any messages
- when cause for the outline does not work in theia
  use the same as the ivy project explorer


ivy-lmu (approved):
👍

ivy-cst (approved):
:+1:

PR #75, opened 2023-09-13
XIVY-12255 Fix select embedded element, select outline correctly (deep)

ivy-lli:
If you selected an element in an embedded process, the element wasn't selected in the process outline.
Now it will be searched deeply and will be correctly selected


ivy-lmu (approved):
👍

PR #90, opened 2023-10-10
XIVY-12449 add first electron playwright tests & integration build

ivy-lmu:
new build: https://jenkins.ivyteam.io/job/vscode-extensions_integration/


build/integration/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM mcr.microsoft.com/playwright:next
```
ivy-cst, line 1: I would use the same version as in yarn.lock

ivy-cst (approved):
Very nice :+1:. May be we should include a real test project?

playwright/package.json
```diff
@@ -0,0 +1,12 @@
+{
+  "private": true,
+  "name": "@axonivy/vscode-extensions-playwright",
+  "version": "12.2.0",
```
ivy-lli, line 4: 12.2.0?

playwright/tests/page-objects/view.ts
```diff
@@ -0,0 +1,34 @@
...
+import { PageObject } from './page-object';
+
+export interface ViewData {
+  tabSelector: string;
+  viewSelector: string;
+  viewName?: string;
+}
+
+export class View extends PageObject {
+  constructor(private data: ViewData, page: Page) {
```
ivy-lli, line 11: i would also mark this parameters as readonly to make them final.

playwright/tests/engine.test.ts
```diff
@@ -0,0 +1,52 @@
...
+  test('check if embedded engine has started', async ({ window }) => {
+    const outputview = new OutputView(window);
+    await outputview.isTabVisible();
+    await outputview.isActive();
+    await outputview.isViewVisible();
+    await outputview.checkIfEngineStarted();
+  });
+
+  test('check default engine settings', async ({}, testInfo) => {
+    const app = await launchElectronApp(noProjectWorkspacePath, testInfo.title);
```
ivy-lli, line 19: not sure about this own test extension...
Doesn't start this test now the ElectronApp twice? Once in the test and once here?

But I assume that you need it right? So playwright doesn't simply start a browser?

But maybe we can add some custom parameters to our own test method so we can directly add the workspace there so it is correctly started right away?

ivy-lli (approved):
Looks good to me in general, however I have some suggestions (which I don't know if everything is possible :) ):

playwright/package.json
ivy-lmu: ups :D

ivy-lmu:
> Very nice 👍. May be we should include a real test project?

I will add one at the time it is needed

playwright/tests/engine.test.ts
ivy-lmu: > Doesn't start this test now the ElectronApp twice? Once in the test and once here?

no, if `{ window }`is not passed, the test extension does not run. So, it's started once here.

yes, browser does not work :(

> But maybe we can add some custom parameters to our own test method so we can directly add the workspace there so it is correctly started right away?

I tried to do that, but could not find a good solution yet. But I plan to improve it somehow in a followup

PR #91, opened 2023-10-10
XIVY12449 parametrize fixture

ivy-lmu:
& download vscode before testing


ivy-lli (approved):
Looks good as far as I can tell :)

PR #93, opened 2023-10-12
XIVY-12449 add inscription view tests

ivy-lmu:


playwright/tests/page-objects/inscription-view.ts
```diff
@@ -0,0 +1,49 @@
+import { Locator, Page, expect } from '@playwright/test';
+import { ViewData } from './view';
+import { IFrameView } from './iframe-view';
+
+export class InscriptionView extends IFrameView {
+  constructor(page: Page) {
+    const outputViewData: ViewData = {
+      tabSelector: 'li.action-item:has-text("IVY")',
+      viewSelector: 'body > div > div:not([data-parent-flow-to-element-id]) >> visible = true'
```
ivy-lli, line 9: what is this data-parent-flow...?

ivy-lli (approved):

playwright/tests/page-objects/inscription-view.ts
ivy-lmu: that is the reference to the nested `.editor-instance` (vscode) element. only the process editor view has such a reference :D.

PR #102, opened 2023-10-20
XIVY-12427 Add toolbar to internal browser view

ivy-lli:
Add toolbar to browser view:
- Support update of the input field with dev-wf-ui patch https://github.com/axonivy/dev-workflow-ui/pull/252
  - iframe's does not allow any access if the origin of the URLs do not match, so the only way to communicate between is with messages https://developer.mozilla.org/en-US/docs/Web/Security/Same-origin_policy#cross-origin_script_api_access
- Just a simple js file at the moment, we would need to implement an additional webview build to transpile this otherwise

Screen Recording 2023-10-20 at 16 22 54


vscode-extensions/engine/extension/src/browser/ivy-browser-view-provider.ts
```diff
@@ -1,17 +1,33 @@
...
     webviewView.webview.html = this.getWebviewContent(webviewView.webview);
+
+    webviewView.webview.onDidReceiveMessage(e => {
+      switch (e.type) {
+        case 'openExternal':
+          try {
+            const url = vscode.Uri.parse(e.url);
+            vscode.env.openExternal(url);
+          } catch {
+            // Noop
```
ivy-lmu, line 23: remove comment

vscode-extensions/engine/extension/src/browser/media/index.js
```diff
@@ -0,0 +1,106 @@
...
+
+window.addEventListener('message', console.log);
+
+onceDocumentLoaded(() => {
+  setInterval(() => {
+    const iframeFocused = document.activeElement?.tagName === 'IFRAME';
+    document.body.classList.toggle('iframe-focused', iframeFocused);
+  }, 50);
+
+  iframe.addEventListener('load', () => {
```
ivy-lmu, line 48: not needed... remove it?

ivy-lmu (approved):
very cool 👍 & many thanks

ivy-cst (commented):
:rocket:

PR #104, opened 2023-10-23
unify extensions into one

ivy-lmu:


.vscode/launch.json
```diff
@@ -7,20 +7,13 @@
...
-        "${workspaceFolder}/vscode-extensions/config-editor/extension/lib/**/*.js",
-        "${workspaceFolder}/vscode-extensions/inscription/extension/lib/**/*.js",
-        "${workspaceFolder}/vscode-extensions/process-editor/extension/dist/**/*.js",
-        "${workspaceFolder}/vscode-extensions/engine/extension/dist/**/*.js",
-        "${workspaceFolder}/vscode-extensions/project-explorer/extension/dist/**/*.js"
+        "${workspaceFolder}/dist/**/*.js",
       ],
-      "sourceMaps": true
+      "sourceMaps": true,
+      "preLaunchTask": "clean & build node"
```
ivy-lmu, line 16: before each start the node part is built quickly 🤩

ivy-cst (approved):
Nice one :+1:   Please @ivy-lli  look closer ;)!
Just css and scripts looks strange in top level, but this was already like this before the change.
image

src/base/commands.ts
ivy-lli, line 1: we could help our devs here with a little bit of ts magic, e.g. by provide a union with all known cammands and let the editor provide an autocomplete for it (https://tsplay.dev/wjp0Ym)

ivy-lli (approved):
Looks much cleaner in my eyes 👍

src/base/commands.ts
ivy-lmu: thanks for the hint, will be considered in a followup

PR #110, opened 2023-10-26
XIVY-11752 create new process command

ivy-lmu:


ivy-cst (approved):
Nice :+1:  May be we could also introduce commands where the process kind is already clear and must not be chosen? Like Create Webservice Process?

PR #125, opened 2023-11-07
Handle all messages, hide model loading

ivy-lli:
image
as long as no severy `NONE` messages appears


ivy-cst (approved):
Nice one, thanks :+1:

PR #129, opened 2023-11-15
XIVY-12783 handle commands as types

ivy-lmu:


ivy-lli (approved):
I like this change a lot ❤️

PR #131, opened 2023-11-15
XIVY-12783 delete project request

ivy-lmu:


src/engine/api/engine-api.ts
```diff
@@ -80,6 +84,12 @@ export class IvyEngineApi {
     await this.runPostRequest(newProjectParams, CREATE_PROJECT_REQUEST);
   }
 
+  public async deleteProject(projectDir: string): Promise<any> {
+    const params = { projectDir };
+    const url = this.projectRequestURL(DELETE_PROJECT_REQUEST.sourcePath).toString();
+    return new Promise(resolve => axios.delete(url, { params, headers, auth }).then(response => resolve(response.data)));
```
ivy-lli, line 90: why is this wrapping in a new promise needed?
something like the following should also work I think, or not?
`axios.delete(url, { params, headers, auth }).then(response => response.data)`

src/engine/api/request.ts
```diff
@@ -9,12 +10,16 @@ export const ACTIVATE_PROJECTS_REQUEST: ProjectRequest = { sourcePath: 'activate
...
 
 export const PROJECT_REQUEST_OPTIONS: http.RequestOptions = {
   auth: 'Developer:Developer',
   method: 'GET'
 };
 
+export const headers = { 'X-Requested-By': 'web-ide' };
+export const auth: AxiosBasicCredentials = { username: 'Developer', password: 'Developer' };
+
 export function makeGetRequest(url: URL, options: http.RequestOptions): Promise<string> {
```
ivy-lli, line 23: I think I like the idea of having a wrapper around the axios or fetch or whatever lib/api that makes requests and automatically add the correct headers and auth.
Then we woundn't need to export those here and so one...

ivy-lli (approved):
Looks good, consider to continue with this wrapper api module

src/engine/api/engine-api.ts
ivy-lmu: thanks, you are right, this promise is useless

src/engine/api/request.ts
ivy-lmu: Yes, I think I'll stick to it when I switch to axios completely

PR #133, opened 2023-11-17
XIVY-12783 exlcude pattern for project detection

ivy-lmu:
& remove ivy container

this configuration name change will break code.ivyteam

### DIFF
--- a/package.json
+++ b/package.json
@@ -215,7 +215,7 @@
           },
           "project.excludePattern": {
             "type": "string",
-            "default": "",
+            "default": "**/target/**",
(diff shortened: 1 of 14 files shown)


package.json
```diff
@@ -197,31 +197,38 @@
...
+          },
+          "engine.url": {
+            "type": "string",
+            "default": "http://localhost:8080/",
+            "description": "Url of the Axon Ivy Engine that will be used.",
+            "order": 1
+          },
+          "project.excludePattern": {
+            "type": "string",
+            "default": "",
```
ivy-cst, line 218: as default you could use something like `**/target/` or how would the pattern look like if we do not want to look into target folders?

ivy-cst (approved):
Nice one :+1:

package.json
ivy-lmu: `**/target/**`would it be

PR #134, opened 2023-11-17
XIVY-12783 deploy p.json on change

ivy-lmu:


ivy-cst (approved):
Nice :+1:

ivy-cst:
Would be cool if it is also deployed on the first time then?

PR #147, opened 2023-11-23
Let raise script fail if one command fails

ivy-cst:


ivy-lmu (approved):

ivy-cst:
@alexsuter we should add this on all raise scripts ...

alexsuter:
> @alexsuter we should add this on all raise scripts ...

can you do that? thank you very much for your effort

PR #171, opened 2023-12-07
XIVY-12892 minor preps for publishing to market

ivy-lmu:


extension/package.json
```diff
@@ -1,9 +1,9 @@
 {
-  "name": "ivy-extension",
+  "name": "designer",
```
ivy-cst, line 2: :smile:

PR #174, opened 2023-12-12
XIVY-12892 release to market via pipeline

ivy-lmu:


ivy-cst (approved):
Looks good :+1: 

In the end you should look that the version number is also updated on the version raise scripts

ivy-lmu:
> Looks good 👍
> 
> In the end you should look that the version number is also updated on the version raise scripts

thanks, I have changed it to `${npm_package_version:0:4}`

extension/package.json
```diff
@@ -22,6 +22,7 @@
     "lint": "eslint --ext .ts,.tsx ./src",
     "package": "vsce package --yarn",
     "package:with:engine": "yarn package -o ${npm_package_name}-with-engine-${npm_package_version}.vsix",
```
ivy-cst, line 24: May you have also do adjust this one?

ivy-cst: Yes, true. But may be you could still change the version inside in the pom? So we see the exact version if we install it this way?

ivy-lmu: changed that, now the output is designer-11.vsix and inside pom the version is 11.3.2023xxxx

PR #192, opened 2024-01-12
XIVY-13138 establish inscription and ivyscript ws connection in node backend

ivy-lmu:
`WebSocketForwarder` inspired by https://github.com/eclipse-glsp/glsp-theia-integration/blob/master/packages/theia-integration/src/node/websocket-connection-forwarder.ts


extension/webviews/process-editor/src/ivy-diagram-widget.ts
```diff
@@ -36,15 +39,33 @@ export abstract class IvyGLSPDiagramWidget extends GLSPDiagramWidget {
...
+
+  private wrapWebSocket(notificationType: NotificationType<string>): IWebSocket {
+    return {
+      send: content => this.messenger.sendNotification(notificationType, HOST_EXTENSION, content),
+      onMessage: callback => this.messenger.onNotification(notificationType, callback),
+      onError: () => {},
+      onClose: () => {},
+      dispose: () => {}
+    };
+  }
```
ivy-lli, line 70: not really a IWebSocket 🙈

ivy-lli (approved):
Looks very easy 🎩. 
I'm only not sure about faking a IWebSocket object 😅

ivy-lmu:
@martin-fleck-at, you might find time to check whether this WebSocket forwarding solution is sufficient for our case. Many thanks

extension/webviews/process-editor/src/ivy-diagram-widget.ts
```diff
@@ -36,15 +39,33 @@ export abstract class IvyGLSPDiagramWidget extends GLSPDiagramWidget {
...
   }
 
-  private initServer(server: string) {
+  private initConnection() {
     if (this.actionDispatcher instanceof GLSPActionDispatcher) {
-      this.actionDispatcher
-        .onceModelInitialized()
-        .finally(() => this.actionDispatcher.dispatch(EnableInscriptionAction.create({ connection: { server } })));
+      this.actionDispatcher.onceModelInitialized().finally(() => {
+        const ivyScript = {
```
martin-fleck-at, line 49: Could you elaborate on the two MessageConnections that you are basically creating here? Once the model is ready, we create those connections to the messenger and expose them through the `EnableInscriptionAction`, however how is that used and why do we need to use WebSocketMessageReader and WebSocketMessageWriter as we only use it as a wrapper for the webview messenger?

Btw, you could probably also call `createWebSocketConnection(IWebSocket)` for a shorthand and "re-use" the created IWebSocket you create per type.

extension/webviews/process-editor/src/ivy-diagram-widget.ts
```diff
@@ -36,15 +39,33 @@ export abstract class IvyGLSPDiagramWidget extends GLSPDiagramWidget {
...
+        const inscription = {
+          reader: new WebSocketMessageReader(this.wrapWebSocket(InscriptionWebSocketMessage)),
+          writer: new WebSocketMessageWriter(this.wrapWebSocket(InscriptionWebSocketMessage))
+        };
+        this.actionDispatcher.dispatch(EnableInscriptionAction.create({ connection: { ivyScript, inscription } }));
+      });
     }
   }
+
+  private wrapWebSocket(notificationType: NotificationType<string>): IWebSocket {
```
martin-fleck-at, line 62: I think it might be a bit cleaner to actually write our own `MessageReader` and `MessageWriter` class instead of mimicking a WebSocket that isn't one but really it is only the naming that makes it confusing so I can understand if we keep it like this. Again, I'm not entirely sure I see who is using the resulting MessageConnection.

martin-fleck-at (changes requested):
In general the code looks fine. Unfortunately I cannot really test it as I don't have access to the custom npm registry. Nevertheless, purely from the code it should probably work. However, I'm not too happy with some of the mimicking as I believe it makes the code harder to understand than it needs to be, see my inline comments.

extension/webviews/process-editor/src/ivy-diagram-widget.ts
ivy-lmu: Thank you for the review. As mentioned before, you should now be able to build and run the extension.

One Connection is used to to start the MonacoLangueClient, see https://github.com/axonivy/inscription-client/blob/12c19bd9801c8dee7996f561fa9d8b4da7e32e91/packages/core/src/ivy-script-client.ts#L12C30-L12C30

and the other one to start the InscriptionClientJsonRpc, see https://github.com/axonivy/inscription-client/blob/12c19bd9801c8dee7996f561fa9d8b4da7e32e91/packages/core/src/inscription-client-jsonrpc.ts#L84

extension/webviews/process-editor/src/ivy-diagram-widget.ts
ivy-lmu: Good point, I will provide a MessageReader and Writer.

yarn.lock
```diff
@@ -2991,6 +2991,13 @@
   resolved "https://registry.yarnpkg.com/@types/vscode/-/vscode-1.85.0.tgz#46beb07f0f626665b52d1e2294382b2bc63b602e"
   integrity sha512-CF/RBon/GXwdfmnjZj0WTUMZN5H6YITOfBCP4iEZlOtVQXuzw6t7Le7+cR+7JzdMrnlm7Mfp49Oj2TuSXIWo3g==
 
+"@types/ws@^8.5.10":
```
ivy-cst, line 2994: Please replace all `https://npmjs-registry.ivyteam.io` with `https://registry.yarnpkg.com` in this file. Thanks.

extension/webviews/process-editor/src/message-reader.ts
```diff
@@ -0,0 +1,76 @@
+import { Disposable, DataCallback, AbstractMessageReader, MessageReader } from 'vscode-jsonrpc';
+import { NotificationType } from 'vscode-messenger-common';
+import { Messenger } from 'vscode-messenger-webview';
+
+export class IvyMessageReader extends AbstractMessageReader implements MessageReader {
```
ivy-lli, line 5: I would name this WebViewMessageReader or something like that...

extension/webviews/process-editor/src/message-writer.ts
```diff
@@ -0,0 +1,23 @@
+import { Message, AbstractMessageWriter, MessageWriter } from 'vscode-jsonrpc';
+import { HOST_EXTENSION, NotificationType } from 'vscode-messenger-common';
+import { Messenger } from 'vscode-messenger-webview';
+
+export class IvyMessageWriter extends AbstractMessageWriter implements MessageWriter {
```
ivy-lli, line 5: Same here -> WebViewMessageWriter

ivy-lli (approved):
Nice work!
Is it now possible to debug this messages via this vscode extension? https://marketplace.visualstudio.com/items?itemName=typefox.vscode-messenger-devtools

ivy-lmu:
> Nice work! Is it now possible to debug this messages via this vscode extension? https://marketplace.visualstudio.com/items?itemName=typefox.vscode-messenger-devtools

I see the handlers in the debug view but not the events. I will check that ;)

PR #198, opened 2024-01-18
XIVY-13143 handle new process and new hd actions triggered by inscription

ivy-lmu:
depends on https://github.com/axonivy/core/pull/4801

After creating a process/hd via inscription, the configuration of the call target is not updated correctly. It seem to be a timing issue, I will check that.

### DIFF
--- a/extension/src/project-explorer/ivy-project-explorer.ts
+++ b/extension/src/project-explorer/ivy-project-explorer.ts
@@ -35,11 +35,12 @@ export class IvyProjectExplorer {
     registerCmd(`${VIEW_ID}.buildProject`, (selection: TreeSelection) => this.execute('engine.buildProject', selection));
     registerCmd(`${VIEW_ID}.deployProject`, (selection: TreeSelection) => this.execute('engine.deployProject', selection));
     registerCmd(`${VIEW_ID}.buildAndDeployProject`, (selection: TreeSelection) => this.execute('engine.buildAndDeployProject', selection));
-    registerCmd(`${VIEW_ID}.addBusinessProcess`, (selection: TreeSelection) => this.addProcess('Business Process', selection));
-    registerCmd(`${VIEW_ID}.addCallableSubProcess`, (selection: TreeSelection) => this.addProcess('Callable Sub Process', selection));
-    registerCmd(`${VIEW_ID}.addWebServiceProcess`, (selection: TreeSelection) => this.addProcess('Web Service Process', selection));
+    registerCmd(`${VIEW_ID}.addProcess`, (selection: TreeSelection, pid: string) => this.addProcess(selection, undefined, pid));
+    registerCmd(`${VIEW_ID}.addBusinessProcess`, (selection: TreeSelection) => this.addProcess(selection, 'Business Process'));
+    registerCmd(`${VIEW_ID}.addCallableSubProcess`, (selection: TreeSelection) => this.addProcess(selection, 'Callable Sub Process'));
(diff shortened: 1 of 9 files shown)


extension/src/project-explorer/ivy-project-explorer.ts
```diff
@@ -35,11 +35,12 @@ export class IvyProjectExplorer {
     registerCmd(`${VIEW_ID}.buildProject`, (selection: TreeSelection) => this.execute('engine.buildProject', selection));
     registerCmd(`${VIEW_ID}.deployProject`, (selection: TreeSelection) => this.execute('engine.deployProject', selection));
     registerCmd(`${VIEW_ID}.buildAndDeployProject`, (selection: TreeSelection) => this.execute('engine.buildAndDeployProject', selection));
-    registerCmd(`${VIEW_ID}.addBusinessProcess`, (selection: TreeSelection) => this.addProcess('Business Process', selection));
-    registerCmd(`${VIEW_ID}.addCallableSubProcess`, (selection: TreeSelection) => this.addProcess('Callable Sub Process', selection));
-    registerCmd(`${VIEW_ID}.addWebServiceProcess`, (selection: TreeSelection) => this.addProcess('Web Service Process', selection));
+    registerCmd(`${VIEW_ID}.addProcess`, (selection: TreeSelection, pid: string) => this.addProcess(selection, undefined, pid));
```
ivy-lli, line 38: will this now also appear on the vscode command palette? because it would be nice IMO if this is a hidden cmd. because it is only triggered by the code

extension/src/websocket-forwarder.ts
```diff
@@ -33,6 +35,22 @@ export class WebSocketForwarder implements vscode.Disposable {
     });
   }
 
+  private handleClientMessage(message: string) {
+    const obj = JSON.parse(message);
```
ivy-lli, line 39: same way as glsp does it?
Don't get me wrong, nothing wrong here, but maybe JSON.parse can be a performance leak here, as it is (as far as I know) not ultimative performant (but maybe also only with really big data, which at least in this direction should never be a problem)

ivy-lli (approved):

extension/src/project-explorer/ivy-project-explorer.ts
ivy-lmu: no, as the command is not listed in package.json under contributions.

extension/src/websocket-forwarder.ts
```diff
@@ -35,16 +35,26 @@ export class WebSocketForwarder implements vscode.Disposable {
     });
   }
 
-  private handleClientMessage(message: string) {
-    const obj = JSON.parse(message);
-    if (obj?.method === 'action') {
-      const handler = this.actionHandlerFor(obj.params);
+  private handleClientMessage(message: unknown) {
+    if (this.isAction(message)) {
```
ivy-lmu, line 39: @ivy-lli feels better, but I can't prove whether it's actually better

extension/src/websocket-forwarder.ts
```diff
@@ -35,16 +35,26 @@ export class WebSocketForwarder implements vscode.Disposable {
     });
   }
 
-  private handleClientMessage(message: string) {
-    const obj = JSON.parse(message);
```
ivy-lli: ah not even needed?

ivy-lmu: nop, if no stringify is run under client writer

extension/src/websocket-forwarder.ts
ivy-lli: Looks good

PR #204, opened 2024-02-01
XIVY-13138 tests for creating hd and call sub via inscription

ivy-lmu:
& remove test project that is not built, as it leads to ugly engine errors during testing.
& another attempt to stabilize process execution tests under Linux


playwright/tests/create-process.test.ts
```diff
@@ -38,23 +37,27 @@ test.describe('Create Process', () => {
...
     const start = processEditor.locatorForElementType('g.start\\:requestStart');
     const end = processEditor.locatorForElementType('g.end\\:taskEnd');
     await processEditor.startProcessAndAssertExecuted(start, end);
   });
 
   test('Assert that process gets redeployed after editing', async () => {
+    await processEditor.hasNoStatusMessage();
     await explorer.addProcess(projectName, processName, 'Business Process');
-    await explorer.hasNoStatusMessage();
+    await explorer.hasStatusMessage('Finished: Deploy Ivy Projects');
```
ivy-lli, line 52: maybe this status message check should be part of the addProcess method?

ivy-lli: or at least the explorer should have a method that do this stuff, instead of duplicate the message here x times and if we change it in the future we need to change it everywhere.

playwright/tests/page-objects/inscription-view.ts
```diff
@@ -22,6 +22,10 @@ export class InscriptionView extends PageObject {
     return this.parent.getByLabel(label, { exact: true });
   }
 
+  async click(label: string) {
```
ivy-lli, line 25: -> clickBtn

ivy-lli (approved):

ivy-lli (approved):

PR #210, opened 2024-02-22
Fix warnings and errors on diagram opening

martin-fleck-at:
- Provide empty implementation of context menu service
- Do not re-use WebviewReadyNotification from GLSP but use custom one


ivy-lmu (approved):
thanks @martin-fleck-at for the fix

PR #217, opened 2024-03-06
XIVY-12913: Improve open-page action + playwright tests

ivy-edp:
I have improved the openPage action. Now it's also possible to simply enter the relative path of a file within the project itself and open it (similar behavior to the old Inscription view). 

I've written two Playwright tests for the openPage-Action:
- One checks if a file is opened (using breadcrumbs). It's a bit of a hack but the quickest way I've found to check the active file.
- The other test verifies if a specific URL is opened in the integrated browser.

Note: Attempting to open external URLs in the integrated browser is blocked. You can still open them through the button on the topright in the integrated browser. The help pages work and can be opened in the integrated browser, which is the most important aspect.

example3_pageOpen

### DIFF
--- a/extension/src/inscription-view/open-page.ts
+++ b/extension/src/inscription-view/open-page.ts
@@ -3,17 +3,16 @@ import * as fs from 'fs';
 import * as path from 'path';
 import { InscriptionActionArgs } from '@axonivy/inscription-protocol';
 import { InscriptionActionHandler } from '../inscription-action-handler';
+import { executeCommand } from '../base/commands';
 
 export class OpenPageActionHandler implements InscriptionActionHandler {
   actionId = 'openPage' as const;
   async handle(actionArgs: InscriptionActionArgs): Promise<void> {
     const path = actionArgs.payload.toString();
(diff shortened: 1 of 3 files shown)


extension/src/inscription-view/open-page.ts
```diff
@@ -22,20 +21,49 @@ function isUrl(absolutePath: string) {
...
+function openInExplorer(absolutePath: string | null) {
+  if (absolutePath) {
+    vscode.commands.executeCommand('vscode.open', vscode.Uri.file(absolutePath));
+  } else {
+    vscode.window.showInformationMessage('The entered url is not valid.');
   }
+}
 
-  return false;
+function getProjectFolder() {
```
ivy-lmu, line 47: there is a command `ivyProjects.getIvyProjects` that gives you the available ivy project paths. I think you can simply check if activeTab.input starts with one of them.

ivy-lmu (approved):
Nice work, thank you 👍 a minor remark on the project resolution, which could be improved, see comment

PR #218, opened 2024-03-07
Fix monaco api changes

ivy-lli:


ivy-lli:
Hi @martin-fleck-at 
I also quickly fixed the vscode integration that it can handle the API changes from the monaco.
But I didn't split the monaco in its own chunck. I think do so will take here more effort...

ivy-lmu (approved):

PR #221, opened 2024-03-21
Align VS Code process editor with improved structure

martin-fleck-at:
- Adapt to new structure with ivy-tool-bar being outside of diagram
- Read HTML file from source and adapt instead of re-creating it
- Slightly improve editor worker location handling in webview


martin-fleck-at:
@ivy-lli With this PR the VS Code extension also makes sure that it uses the new structure for adapted UI extensions.

ivy-lli (approved):
Looks good to me

ivy-lmu (approved):
👍

PR #233, opened 2024-04-15
XIVY-13807 build form: modify method saveData to saveDataAndBuild

ivy-lmu:
& update file watcher so that `*.f.json` changes are ignored (negation is not supported...)

### DIFF
--- a/extension/src/editors/form-editor/webview-communication.ts
+++ b/extension/src/editors/form-editor/webview-communication.ts
@@ -15,7 +15,7 @@ export const setupCommunication = (messenger: Messenger, webviewPanel: vscode.We
   }
   const messageParticipant = messenger.registerWebviewPanel(webviewPanel);
   const toDispose = new DisposableCollection(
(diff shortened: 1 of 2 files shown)


extension/src/editors/form-editor/webview-communication.ts
```diff
@@ -24,3 +24,16 @@ export const setupCommunication = (messenger: Messenger, webviewPanel: vscode.We
...
+    if (this.isSaveData(message)) {
+      message.method = 'saveDataAndBuild';
+    }
+    super.handleClientMessage(message);
+  }
+
+  isSaveData = (obj: unknown): obj is { method: string } => {
+    return typeof obj === 'object' && obj !== null && 'method' in obj && obj.method === 'saveData';
+  };
+}
```
ivy-lli, line 39: with this you may can even implement the dirtyflag mode.
1. By catch the saveData message and hold the payload 
2. Set the dirty state to true
3. On real editor save (by the user ctrl + s) -> send saveDataAndBuild with the last saveData payload

ivy-lli (approved):

PR #237, opened 2024-04-18
XIVY-13804 support add new form

ivy-lmu:


extension/package.json
```diff
@@ -204,6 +204,11 @@
         "title": "New Html Dialog",
         "category": "Axon Ivy"
       },
+      {
+        "command": "ivyProjects.addNewForm",
+        "title": "New Form",
```
ivy-rew, line 209: This is visible to the user right? -> I'd call it 'New Dialog Form' ... so that the Dialog concept is linkable to  the other dialog natures.

ivy-lmu: yes it's visible, I would then streamline to "New Form Dialog"

extension/package.json
```diff
@@ -204,6 +204,11 @@
         "title": "New Html Dialog",
         "category": "Axon Ivy"
       },
+      {
+        "command": "ivyProjects.addNewForm",
```
ivy-rew, line 208: addNewFormDialog?

ivy-rew (approved):

PR #240, opened 2024-04-22
XIVY-11980 integrate variables editor

ivy-lgi:


playwright/tests/page-objects/variables-editor.ts
```diff
@@ -12,43 +12,30 @@ export class VariablesEditor extends Editor {
...
 
   async hasKey(key: string) {
-    const field = this.viewFrameLoactor().locator(`vscode-text-field[grid-column="1"][current-value="${key}"]`);
+    const field = this.viewFrameLoactor().locator(`//td/div/span[contains(text(), '${key}')]`);
     await expect(field).toBeVisible();
   }
 
   async hasValue(value: string) {
-    const field = this.viewFrameLoactor().locator(`vscode-text-field[grid-column="2"][current-value="${value}"]`);
+    const field = this.viewFrameLoactor().locator(`//td/div[contains(text(), '${value}')]`);
```
ivy-lli, line 27: not such a big fan of xpath locators...
I think it should also be possible to write this with CSS selectors

ivy-lli (approved):
Looks good. 
except the xpath stuff 🙈

playwright/tests/page-objects/page-object.ts
```diff
@@ -3,14 +3,18 @@ import { Locator, Page, expect } from '@playwright/test';
 export class PageObject {
   constructor(readonly page: Page) {}
 
-  async executeCommand(command: string) {
+  async executeCommand(command: string, ...options: Array<string>) {
```
ivy-lmu, line 6: is there a use case for more than one option?

ivy-lmu (approved):

playwright/tests/page-objects/page-object.ts
ivy-lgi: Yes, with an improvement I will make in the next commit, where I simply call `this.provideUserInput(option);` in the loop over `options`, something like:
https://github.com/axonivy/vscode-extensions/blob/38ecdbfc585e4396ae31e73f2a26df735b02d18b/playwright/tests/page-objects/explorer-view.ts#L76-L80

can be replaced with:
`await this.executeCommand('Axon Ivy: New Project', projectName, '', '', '');`

ivy-lmu: very good 👍

ivy-lli (approved):

PR #246, opened 2024-04-25
XIVY-13804 form creation via inscription

ivy-lmu:


extension/src/editors/process-editor/inscription-view/new-user-dialog.ts
```diff
@@ -2,15 +2,32 @@ import * as vscode from 'vscode';
...
+    }
+    const command = dialogType === 'Form' ? 'ivyProjects.addNewFormDialog' : 'ivyProjects.addNewHtmlDialog';
+    await executeCommand(command, tabInput, [], actionArgs.context.pid);
+    sendInscriptionNotification('dataChanged');
+    sendInscriptionNotification('validation');
+  }
+
+  private collectDialogType(): Promise<DialogType | undefined> {
+    return vscode.window.showQuickPick(
+      dialogTypes.filter(t => t !== 'JSFOffline'),
```
ivy-rew, line 26: can we filter only if the caller is a UserTask?

ivy-lmu: the current action payload does not contain this information.

ivy-rew: let's drop the filter than ... and just allow any view-technology to be used on dialog creation.

ivy-rew (approved):

PR #255, opened 2024-05-08
Raise dependencies version to 11.4.0-SNAPSHOT

ivy-team:


.ivy/raise-deps.sh
```diff
@@ -3,7 +3,9 @@ set -e
 
 shopt -s globstar
 # likely not working on mac
-sed -i -E "s|core_product-engine/job/master/|core_product-engine/job/release%2F${1/.[0-9]-SNAPSHOT/}/|" build/**/Jenkinsfile
+
+# This should only be done if branch is not master...
+# sed -i -E "s|core_product-engine/job/master/|core_product-engine/job/release%2F${1/.[0-9]-SNAPSHOT/}/|" build/**/Jenkinsfile
```
ivy-lli, line 8: This should only be done if the current branch is not the master branch IMO.
Otherwise it will replace job/master with job/release11.4 which not exists for the master branch...
@ivy-cst maybe we need to expose the other parameters too from the raise job?

ivy-lli: hm... yes sounds right to have a create-release branch script where this should be included.

PR #256, opened 2024-05-10
Fix builds for newest artifacts/versions

ivy-lli:
- Raise engine version to v11.4.0
- Fix Variables / Form editor tests
- Add raise-build-plugin-version script
- Improve raise-version script, so it also raises the ivy engine version in maven.ts

### DIFF
--- a/playwright/tests/form-editor.test.ts
+++ b/playwright/tests/form-editor.test.ts
@@ -33,7 +33,7 @@ test.describe('Form Editor', () => {
   test('Edit input label', async () => {
     const input = editor.locatorFor('.block-input');
     await input.click();
-    const labelProperty = editor.locatorFor('[id=":re:-input"]');
+    const labelProperty = editor.locatorFor('#properties .ui-input').first();
     const newLabel = randomArtefactName();
     await labelProperty.fill(newLabel);
     await expect(input).toHaveText(newLabel);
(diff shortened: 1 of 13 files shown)


playwright/tests/page-objects/variables-editor.ts
```diff
@@ -12,8 +12,7 @@ export class VariablesEditor extends Editor {
   }
 
   override async isViewVisible() {
-    const header = this.viewFrameLoactor().locator('.master-toolbar > div > div');
-    expect(header).toHaveText('Variables Editor');
+    const header = this.viewFrameLoactor().locator('.master-toolbar:has-text("Variables Editor")');
```
ivy-lli, line 15: fyi @ivy-lgi: the locator '.master-toolbar > div > div' lead to an error where the selector matched to two results (which cause an error in playwright). And an expect should always have an `await` as it is asynchron.

playwright/tests/form-editor.test.ts
```diff
@@ -33,7 +33,7 @@ test.describe('Form Editor', () => {
   test('Edit input label', async () => {
     const input = editor.locatorFor('.block-input');
     await input.click();
-    const labelProperty = editor.locatorFor('[id=":re:-input"]');
+    const labelProperty = editor.locatorFor('#properties .ui-input').first();
```
ivy-lli, line 36: fyi @ivy-lmu: ":re:" is a random id by react, so not really a good idea to use this in a test case :)

ivy-cst (approved):
Thanks a lot :+1:

ivy-rew (commented):
great work; thanks @ivy-lli

ivy-lmu:
thx 👍

PR #257, opened 2024-05-21
Fix broken monaco test because of completion imrpovement

ivy-rew:


ivy-rew:
open-vscode is green: so ready to integrate :wink: @weissreto

weissreto (approved):

PR #258, opened 2024-06-06
XIVY-14130 provide playwirght-jdk17 docker image

ivy-lmu:


alexsuter (approved):

ivy-cst (commented):
Nice. If it is for build only it would also be an option to move it to the build images

PR #262, opened 2024-06-10
fix variables editor

ivy-lgi:
after implementing communication with the back end to get validation messages, we need to wait for the engine to start before setting up the connection


ivy-lmu (approved):
looks good, thanks 👍

PR #274, opened 2024-06-24
XIVY-14102 remove unnecessary build & deploy, engine does it now

ivy-lmu:


ivy-lmu:
I merge it, because there also some glsp issues that break tests. So Lukas can continue ;)

ivy-cst:
:+1:

PR #277, opened 2024-07-01
refactor: cleanup internal command ussage & process.env

ivy-lmu:
use API instead of commands and process.env


extension/src/browser/ivy-browser-view-provider.ts
```diff
@@ -24,13 +44,27 @@ export class IvyBrowserViewProvider implements vscode.WebviewViewProvider {
           }
           break;
         case 'openHome':
-          executeCommand('engine.openDevWfUi');
+          IvyEngineManager.instance.openDevWfUi();
```
ivy-lli, line 47: not sure, but maybe this functions on the IvyEngineManager (openDevWfUi, openEngineCockpit, etc), should be part of this provider?
Yes maybe the base needs to come from the engine manager, but the functionality should deliver this one here right?

Because now you call here the IvyEngineManager and this will call this manager here again.
Or did I read this code wrong?

ivy-lli (approved):
Looks like the total right direction!
Internally use fully typed functions, provide commands which can be triggered why the vscode command palette. Nice :)

PR #280, opened 2024-07-02
let the engine run alongside during testing

ivy-lmu:
in most tests the embedded engine is no longer started, instead an engine runs alongside
still the embedded engine gets tested, e.g. with `Create Project` test

- adapt tests and pipelines to make this work
- use of a status bar item instead of a message to indicate that the extension has been started


ivy-lli (approved):
Are you a 🧙‍♂️?

PR #282, opened 2024-07-03
run var tests

ivy-lmu:


ivy-lmu:
@ivy-lgi please fix the commented checks. I would recommend designing the test so that it is repeatable on failure.

PR #288, opened 2024-07-11
XIVY-12909 axios client codegen with orval

ivy-lmu:
- script for downloading openapi.yaml files from given engien url
- codegen axios client & adapt to new clients


orval.config.ts
```diff
@@ -0,0 +1,18 @@
+import { defineConfig } from 'orval';
+
+export default defineConfig({
```
ivy-lmu, line 3: per default axios client is used. here we can also define client fetch for neo.

ivy-lli, line 3: maybe we could think about a filter, for only those api's we really use here? 
https://orval.dev/reference/configuration/input#tags

ivy-lli (approved):
Cool. Only thing which is missing I think is a build step where it is checked if the protocol is still compatible

ivy-lmu:
> Cool. Only thing which is missing I think is a build step where it is checked if the protocol is still compatible

yes, will do it in a followup

orval.config.ts
ivy-lmu: good one, I will check

orval.config.ts
```diff
@@ -0,0 +1,23 @@
+import { defineConfig } from 'orval';
+
+const hooks = { afterAllFilesWrite: 'prettier --write' };
+const filters = { tags: ['web-ide'] };
```
ivy-lli, line 4: 👍

ivy-lli (approved):

PR #292, opened 2024-07-15
use tsc for protocol check

ivy-lmu:


extension/package.json
```diff
@@ -24,6 +24,7 @@
     "package:with:engine": "bash -c 'vsce package -o ${npm_package_name}-with-engine.vsix --no-update-package-json --no-git-tag-version --yarn ${npm_package_version:0:4}.$(date \"+%Y%m%d%H\")'",
     "publish:next": "bash -c 'vsce publish --yarn --no-update-package-json --no-git-tag-version ${npm_package_version:0:4}.$(date \"+%Y%m%d%H\")'",
     "prepare": "yarn clean && yarn build",
+    "type": "tsc",
```
ivy-lli, line 27: good, but make sure that the noEmit flag is enabled in the tsconfig! otherwise tsc will also transpile the code to js what is not needed for type checking.

ivy-lli (approved):

extension/package.json
ivy-lmu: it's not set and even with noEmit=false I do not get .js output

ivy-lli: delete the tsconfig.tsbuildinfo ;)

ivy-lli:
Instead of add it to each tsc command you could also added it to the tsconfig.base.json

ivy-lmu:
> Instead of add it to each tsc command you could also added it to the tsconfig.base.json

build is going to fail:
ERROR in ./src/index.ts
Module build failed (from ../node_modules/ts-loader/index.js):
Error: TypeScript emitted no output for /Users/lmu/git/vscode-extensions/extension/src/index.ts.
    at makeSourceMapAndFinish (/Users/lmu/git/vscode-extensions/node_modules/ts-loader/dist/index.js:55:18)
    at successLoader (/Users/lmu/git/vscode-extensions/node_modules/ts-loader/dist/index.js:42:5)
    at Object.loader (/Users/lmu/git/vscode-extensions/node_modules/ts-loader/dist/index.js:23:5)

PR #320, opened 2024-09-02
XIVY-14827 Remove process outline

ivy-lli:
- Now included into the process editor

Screenshot 2024-09-02 at 13 09 40


ivy-lmu (approved):
👍 
I think you can also remove the `ProcessOutline.p.json`from the test project.
I assume that the outline is tested in the inscription repo and therefore no test is needed here.

PR #326, opened 2024-09-09
fix jsonrpc imports

ivy-lmu:


ivy-lli (approved):
Thx for the fix
Changed that the jsonrpc now gets built with vite, so this imports should now work directly instead over this lib stuff

ivy-lmu:
> Thx for the fix Changed that the jsonrpc now gets built with vite, so this imports should now work directly instead over this lib stuff

I like :)

PR #343, opened 2024-10-29
fix data mapping and console log output check

ivy-lmu:


playwright/tests/page-objects/output-view.ts
```diff
@@ -12,7 +12,7 @@ export class OutputView extends View {
   }
 
   async checkIfEngineStarted() {
-    const expectedText = "Type 'shutdown' and confirm";
+    const expectedText = 'Axon Ivy Engine is running and ready to serve.';
```
ivy-lmu, line 15: 🥑

ivy-cst: Ups :+1:

ivy-cst (approved):

PR #353, opened 2024-11-26
raise orval to 7.3 and node to 22.11

ivy-lmu:


package.json
```diff
@@ -2,7 +2,7 @@
   "name": "root",
   "private": true,
   "engines": {
-    "node": ">=20.10.0"
+    "node": ">=22.11.0"
```
ivy-lli, line 5: IMO this is not really needed. 
Technically it runs fine with node v20 so I'm not sure if we should restrict this here.
Even there are maybe problems with the raise jobs if they still use node v20 then the job can not run anything in this repo :)

ivy-lli (commented):
thx 👍

package.json
ivy-lmu: alright, I'll undo it

PR #355, opened 2024-11-26
raise build-plugin

ivy-lmu:


package-lock.json
```diff
@@ -34,7 +34,7 @@
       }
     },
     "extension": {
-      "name": "designer-11",
+      "name": "designer-13",
```
alexsuter, line 37: why version needed?

alexsuter (changes requested):

package-lock.json
ivy-lmu: in order to make this thing unique

ivy-lmu: see id:
image

PR #382, opened 2025-01-24
Fix build

ivy-lli:
I've struggled lately with fixing the package.lock after updates. 
The main problem was that our editor dependencies were moved from "node_modules/...editor" to "extension/node_modules/...editor". After such a change the build was failing as the dependencies were no longer be found.

It took me some time to realize what the problem was, as nether the root or the extension has a dependency to any editors. 
I think the problem is the npm has problems when you have packages in other packages:
- root
  - extension
    - webviews/... 
As it then don't hoist dependencies up to the root node_modules but to the next upper level, in this case extension.

I think we don't need to have the webview code inside the extension, only the built output. So I moved the webviews folder to the root directory. 
The output of the build step is still inside the extension itself "extension/dist/webviews", so everything should still work as expected


ivy-lmu (approved):
good catch, thanks for fixing

PR #390, opened 2025-01-30
fix build: avoid nested node_modules

ivy-lmu:
move deps that get updated automatically to the root

### DIFF
--- a/extension/package.json
+++ b/extension/package.json
@@ -27,9 +27,6 @@
   },
   "main": "./dist/extension.js",
   "dependencies": {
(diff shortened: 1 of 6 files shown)


ivy-lmu:
> Hmm... seems to be wrong in my eyes. I assume that is because of the bom stuff? Strange that it cannot handle this things.

maybe it's not so wrong. when I add a dependency to a nested package.json for the first time, the dependency is added to the root node_modules, e.g. `"node_modules/@axonivy/form-editor-protocol"`

when this dependency is updated, it gets nested, e.g. ` "webviews/form-editor/node_modules/@axonivy/form-editor-protocol"`

PR #406, opened 2025-02-11
switch linux integration to macos

ivy-lmu:
running playwright tests with xvfb seems to be a bad idea, the screen freezes all the time. So I suggest switching to mac.


build/integration/Jenkinsfile
```diff
@@ -1,5 +1,12 @@
 pipeline {
-  agent any
+  agent {
+    label 'bigSur'
```
alexsuter, line 3: oh no, what are you doing here?

build/integration/Jenkinsfile
```diff
@@ -60,29 +54,28 @@ pipeline {
...
-        withChecks('Integration Protocol ESLint') {
-          recordIssues enabledForFailure: true, publishAllIssues: true, aggregatingResults: true, tools: [esLint(id: 'latest', pattern: 'extension/**/eslint.xml')], qualityGates: [[threshold: 1, type: 'TOTAL', unstable: true]]
         }
       }
     }
   }
 
   post {
+    always {
+      sh 'nohup extension/AxonIvyEngine/bin/AxonIvyEngine stop'
```
alexsuter, line 77: why nohup? I like hup!

alexsuter (changes requested):

alexsuter (approved):
If you aligned these build change with your cüpli-brother (christian). Then I'm good.

ivy-cst:
> running playwright tests with xvfb seems to be a bad idea, the screen freezes all the time. So I suggest switching to mac.

All other stuff is doing well like this. Ok doing it also on mac, but will be much slower and you have only one build slot...

ivy-lmu:
> > running playwright tests with xvfb seems to be a bad idea, the screen freezes all the time. So I suggest switching to mac.
> 
> All other stuff is doing well like this. Ok doing it also on mac, but will be much slower and you have only one build slot...

I will try to run it with github actions, at least for mac

PR #426, opened 2025-02-20
make integration tests run on github

ivy-lmu:
- introduce linux integration pipeline using github actions
- make one test fixture that is always used
- improve download engine script: follow redirects


ivy-lmu:
not yet solved, how do we keep track of the daily build result (scrum-view)

ivy-lli:
Maybe I missed that point, but why should this be more stable on github action than on our jenkins? Or what is the point of moving it here? 🙈

ivy-lmu:
> Maybe I missed that point, but why should this be more stable on github action than on our jenkins? Or what is the point of moving it here? 🙈

1. out of curiosity ^^
2.  on Jenkins it was double virtualized because xvfb was running in a Docker container. My guess is that this double virtualization makes the tests unreliable, but I can't prove it. I just noticed that on github, where xvfb is running directly on the runner, the tests are much more stable.

ivy-lmu:
Of course I could try to directly run xvfb on the Jenkins node without a Docker container. but that does not work out of the box I think.

ivy-lmu:
few words about the one and only test fixture. It is now very easy to set up a temporary workspace for each test and thus create an isolated test environment. But that's the next step ;)

package.json
```diff
@@ -30,15 +30,15 @@
...
-    "@axonivy/process-editor": "13.1.0-next.1358",
-    "@axonivy/process-editor-inscription": "13.1.0-next.1358",
-    "@axonivy/variable-editor": "13.1.0-next.552",
-    "vscode-messenger-common": "0.5.1"
+    "@axonivy/form-editor": "13.1.0-next.578",
+    "@axonivy/form-editor-core": "13.1.0-next.578",
+    "@axonivy/process-editor": "13.1.0-next.1400",
+    "@axonivy/process-editor-inscription": "13.1.0-next.1400",
+    "@axonivy/variable-editor": "13.1.0-next.586",
+    "vscode-messenger-common": "0.4.5"
```
ivy-lli, line 38: why downgrade?

scripts/download-engine.cts
```diff
@@ -13,10 +13,16 @@ function downloadEngine() {
...
-  const filename = path.join(engineDir, path.basename(engineDonwloadUrl));
-
-  https.get(engineDonwloadUrl, res => {
+function downloadEngineReq(engineDir: string, downloadUrl: string) {
+  const filename = path.join(engineDir, path.basename(downloadUrl));
+  https.get(downloadUrl, res => {
+    if (res.statusCode === 302 && res.headers.location) {
+      downloadEngineReq(engineDir, res.headers.location);
+      return;
+    }
```
ivy-lli, line 25: normally this clients have a follow redirect mode and you don't have to do this on your own...

playwright/tests/fixtures/baseTest.ts
```diff
@@ -0,0 +1,77 @@
...
+        test.info().attachments.push({ name: 'trace', path: tracePath, contentType: 'application/zip' });
+        test.info().attachments.push({ name: 'screenshot', path: tracePath, contentType: 'image/png' });
+      }
+      await electronApp.close();
+    }
+  }
+});
+
+const appWithPage = async (workspace: string): Promise<{ page: Page; browser?: Browser; electronApp?: ElectronApplication }> => {
+  if (runInBrowser) {
```
ivy-lli, line 36: a little bit stragne to check this here return browser and the electronApp here as optional and then check it again in the test. why not check it in the test and then do also the intitalize and take for each, maybe even directly in the browser and electron functions.

ivy-lli (approved):
Looks good, some comments
Fine for me to test this stuff on github actions, however I think we need to have a solution that we see when it fails. 
At least we can have the browser version on our jenkins, so we see there if something is broken in general right?

package.json
ivy-lmu: getting trouble with eclipse-glsp they are using 0.4.5

ivy-lmu:
> Looks good, some comments Fine for me to test this stuff on github actions, however I think we need to have a solution that we see when it fails. At least we can have the browser version on our jenkins, so we see there if something is broken in general right?

yes we have the openvscode integration and windows https://jenkins.ivyteam.io/job/vscode-extensions_integration-windows/

scripts/download-engine.cts
ivy-lmu: could indeed be more elegant, but that's not the focus for me, just an help script

playwright/tests/fixtures/baseTest.ts
```diff
@@ -11,44 +11,22 @@ export const runInBrowser = process.env.RUN_IN_BRWOSER ? true : false;
...
-        const screenshotPath = test.info().outputPath('screenshot.png');
-        await page.context().tracing.stop({ path: tracePath });
-        await page.screenshot({ path: screenshotPath });
-        test.info().attachments.push({ name: 'trace', path: tracePath, contentType: 'application/zip' });
-        test.info().attachments.push({ name: 'screenshot', path: tracePath, contentType: 'image/png' });
-      }
-      await electronApp.close();
-    }
+    if (runInBrowser) await runBrowserTest(workspace, take);
+    else await runElectronAppTest(workspace, take);
```
ivy-lli, line 15: I know this is valid syntax, however I'm more used to read the brackets syntax...
```suggestion
    if (runInBrowser) {
      await runBrowserTest(workspace, take);
    } else {
      await runElectronAppTest(workspace, take);
    }
```

ivy-lli (approved):
Yes much more readable I think 👍

ivy-cst:
> > Maybe I missed that point, but why should this be more stable on github action than on our jenkins? Or what is the point of moving it here? 🙈
> 
>     1. out of curiosity ^^
> 
>     2. on Jenkins it was double virtualized because xvfb was running in a Docker container. My guess is that this double virtualization makes the tests unreliable, but I can't prove it. I just noticed that on github, where xvfb is running directly on the runner, the tests are much more stable.

I guess also github uses virtual hosts and inside them containers for the runners ... (but not docker I think)
May be if we had the same software in the container as they do it would may be work?
https://github.com/actions/runner-images/blob/ubuntu24/20250209.1/images/ubuntu/Ubuntu2404-Readme.md

Would be interesting if it would also work with a self hosted runner: https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/about-self-hosted-runners

And what they install on this host ... I am sure they use containers ...

PR #465, opened 2025-03-19
simplify openvscode-server fetcher

ivy-rew:
slightly easier to re-use and copy from ... since we do not restate the version again and again.


ivy-lmu (approved):
thx wget pro 👍

PR #468, opened 2025-03-24
[Demo] Add marker navigation workaround

tortmayr:
Add the marker navigation workaround as proposed in https://github.com/eclipse-glsp/glsp/issues/206#issuecomment-938067519

Application of the proposed workaround for https://github.com/eclipse-glsp/glsp/issues/416
As mentioned in the linked comment, the solution is not ideal and probably needs to be improved on.

### DIFF
--- a/extension/src/editors/process-editor/process-editor-provider.ts
+++ b/extension/src/editors/process-editor/process-editor-provider.ts
@@ -3,12 +3,13 @@ import {
   GlspEditorProvider,
   SocketGlspVscodeServer,
   Writable,
-  configureDefaultCommands
+  configureDefaultCommands,
+  CenterAction
 } from '@eclipse-glsp/vscode-integration';
 import * as vscode from 'vscode';
 import { setupCommunication } from './webview-communication';
 import { createWebViewContent } from '../webview-helper';
-import { ProcessVscodeConnector } from './process-vscode-connector';
+import { DIAGNOSTIC_CLIENT_ID_QUERY_PARAM, DIAGNOSTIC_ELEMENT_ID_QUERY_PARAM, ProcessVscodeConnector } from './process-vscode-connector';
 import { messenger } from '../..';
 
 export default class ProcessEditorProvider extends GlspEditorProvider {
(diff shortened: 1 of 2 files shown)


ivy-lmu:
Cool, thanks for the workaround @tortmayr. I have added a selection action beside the center.

Currently it does not work with editors in preview mode. If the process editor is in preview mode and a problem is selected, the editor is closed. The same applies to other editors that are in preview mode. See:

https://github.com/user-attachments/assets/88bea4d9-fb99-4394-a089-2170770dcebd

Is there any chance of solving this problem?

tortmayr:
@ivy-lmu 
Resolving this issue turns out to be trickier than expected.
Just checking the vscode setting for `workbench.editor.enablePreview' is not enough. We still don't know 
if an open editor is really in preview mode or not. There is no VSCode API to query this from extension side.

The only reliable workaround is to not close the exsiting diagram editor at all when starting a navigation.
I have pushed a commit to change this.
However, this also has some unwanted side effects. Since we now have the same model/diagram open twice there a are sync issues when closing one of the editors and or make changes.
Maybe this is something that can be resolved on server side.

Also, there is currently no way in GLSP to reliably send actions from the extension side once the diagram is initialized (i.e. the Center and Select action). I have worked around this for now using a timeout and opened https://github.com/eclipse-glsp/glsp/issues/1513 to track this.

ivy-lmu:
> We still don't know if an open editor is really in preview mode or not. There is no VSCode API to query this from extension side.

@tortmayr you can find the preview state of the tabs with :
```
    vscode.window.tabGroups.all.forEach(group => {
      group.tabs.forEach(tab => {
        console.log('tab', tab, tab.isPreview);
      });
    });
```
what do you think, would that work?

tortmayr:
Hi @ivy-lmu, I finally had time to look into this.

Unfortunately using the preview state of the tab is not reliable.
This state is not updated correctly if the preview mode is changed via user interaction (e.g. via edit or pinning via double-click ). This seems to be a known issue ([vscode#145734](https://github.com/microsoft/vscode/issues/145734),[vscode#145755](https://github.com/microsoft/vscode/issues/145755) ,[vscode#145773](https://github.com/microsoft/vscode/issues/145773))

However, I came up with a different solution. In the end we don't need to know if an editor tab is in preview or not.
It can be broken down to:
preview mode -> vscode closes currently open editor for the nav uri, then opens a new one
non-preview mode -> currently open editor remains open, second editor for the nav uri is opened

So when handling navigation we basically just need to check if there is another open editor for the same uri (i.e. registered GLSPVscodeClient). If so send the navigation actions to the client for this editor and close the second editor asap. If not, proceed with opening the document normally then send the navigation actions to the newly opened editor.

This solution seems to work great, is stable and so far I don't have encountered any issues.

However, there is still one downside that comes with the new navigation support. Since we have to encode the navigation information as part of the diagnostic URI we end up with a different URI for each problem. This means Vscode can no longer group them accordingly in the problem view:

image

So instead of grouping problems for the same resource under one root node, we will get a new root node for each problem.

There is no way to fix this no other way to transfer the navigation information.

So in the end its a trade-off and you have decide If you want the navigation support even if ti messes with problem view grouping.

tortmayr:
Closes in favor of #632

PR #509, opened 2025-04-16
fix: add dev mode flag

ivy-lmu:
interesting, the flag is not needed when running in the browser (openvscode)


ivy-lli:
Thanks a lot, my hero @ivy-lmu 🙇

PR #532, opened 2025-04-29
Initial support for diagram diffing

tortmayr:
Implement inital (client-side) support for diagram diffing with the process editor

- Update configuration defaults for p.json files so that files associated with a git scheme are opened with the text editor by default -> When opening a diff view for a p.json file from the git view the default text compare editor is openend by default
- Add an entry to the title bar of the text diff editor that allows reopening in the diagram compare view (use artificial uris with pdiff extension so that we can distinguish between files that are openend normally and in compare mode)
- Update process editor provider to handle .pdiff uris
  - Extract diff info (id, side, uri) from pdiff uris forward it to the webview
- Update the webview so that it attaches optionally present diff information to the args of the `RequestModelAction`

Missing:
- Server-side implementation. Currently the optional diff args are not evaluated and the server just openes the file normally

### DIFF
--- a/extension/package.json
+++ b/extension/package.json
@@ -49,6 +49,11 @@
     "onStartupFinished"
   ],
   "contributes": {
+    "configurationDefaults": {
+      "workbench.editorAssociations": {
+        "{git,gitlens}:/**/*.p.json": "default"
+      }
+    },
     "customEditors": [
       {
         "viewType": "ivy.variableEditor",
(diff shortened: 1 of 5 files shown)


ivy-lmu:
Hi @tortmayr, should this pr be reviewed already? I ask because it is in draft state.

tortmayr:
@ivy-lmu Sorry for that. Yes the PR is ready for review.

PR #610, opened 2025-06-12
lock file maintenance and make it reproducible

ivy-lmu:
lock file maintenance procedure:
- npm run update:axonivy:next
- delete package-lock.json
- npm install
- npm ddp

with the fixtures I have made, this procedure should currently be reproducible.

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,6 +54,7 @@
     "@types/react-dom": "19.1.6",
     "@vitejs/plugin-react": "4.5.1",
     "adm-zip": "0.5.16",
+    "globals": "14.0.0",
     "i18next": "^25.0.0",
     "orval": "7.9.0",
     "path": "0.12.7",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -68,7 +69,8 @@
     "vite-tsconfig-paths": "5.1.4"
   },
   "overrides": {
-    "vscode-jsonrpc": "8.2.0"
+    "vscode-jsonrpc": "8.2.0",
+    "@eclipse-glsp/sprotty": "2.3.0"
```
ivy-lmu, line 73: otherwise 2.4.0 will be used which does not work.

package.json
```diff
@@ -54,6 +54,7 @@
     "@types/react-dom": "19.1.6",
     "@vitejs/plugin-react": "4.5.1",
     "adm-zip": "0.5.16",
+    "globals": "14.0.0",
```
ivy-lmu, line 57: need to be fixed here, otherwise eslint does not work.

ivy-lli (approved):

PR #632, opened 2025-06-25
[Demo] Add marker navigation workaround

martin-fleck-at:
Add the marker navigation workaround as proposed in https://github.com/eclipse-glsp/glsp/issues/206#issuecomment-938067519

Application of the proposed workaround for https://github.com/eclipse-glsp/glsp/issues/416
As mentioned in the linked comment, the solution is not ideal and probably needs to be improved on.


ivy-lmu:
Thanks @martin-fleck-at, looks good now 👍

I'm not sure yet whether we should merge this workaround or create a custom problem view. In any case, consider this task completed. Thank you

PR #648, opened 2025-07-10
XIVY-16480 add Runtimelog to vs code extension

ivy-tgr:


package.json
```diff
@@ -35,6 +35,8 @@
     "@axonivy/form-editor": "~13.2.0-next.883.4563925",
     "@axonivy/form-editor-core": "~13.2.0-next.883.4563925",
     "@axonivy/jsonrpc": "~13.2.0-next.782",
+    "@axonivy/log-view": "~13.2.0-next.177.45fd3e3",
```
ivy-lmu, line 38: view itself is not needed I think?

extension/src/engine/ws-client.ts
```diff
@@ -3,8 +3,28 @@ import { WebIdeClientJsonRpc } from './api/jsonrpc';
...
 export const WebSocketClientProvider = (webSocketUrl: URL) => {
+  const runtimeLogWebSocket = new WebSocket(new URL('ivy-runtime-log-lsp', webSocketUrl));
+  runtimeLogWebSocket.onopen = () => {
+    const connection = toSocketConnection(runtimeLogWebSocket);
+    const outputChannel = vscode.window.createOutputChannel('Runtime Log');
+
+    LogClientJsonRpc.startClient(connection).then(client => {
+      client.data().then(data =>
+        data.forEach(entry => {
+          outputChannel.appendLine(`[${entry.timestamp}][${entry.level}][${entry.project}] ${entry.message} ${entry.stacktrace}`);
```
ivy-lmu, line 17: ${entry.stacktrace} can be undefined, don't print if so.

extension/src/engine/ws-client.ts
```diff
@@ -3,8 +3,28 @@ import { WebIdeClientJsonRpc } from './api/jsonrpc';
...
+          outputChannel.show();
```
ivy-lmu, line 18: I would not focus it because we want to focus on dev-wf-ui

ivy-lmu (changes requested):
good one, see comments.

I think you can improve the output in a followup PR. See for example git output:
image

there you can see some coloring stuff and there is also a log level filter. I assume you have to define a log language to achieve this, but we can research...

package-lock.json
```diff
@@ -20139,6 +20158,15 @@
...
     },
+    "webviews/runtimelog-view": {
+      "name": "@axonivy/vscode-webview-log-view",
+      "version": "13.2.0",
+      "extraneous": true,
+      "license": "(EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0)",
+      "dependencies": {
+        "@axonivy/vscode-webview-common": "~13.2.0-next"
+      }
+    },
```
ivy-lmu, line 20169: can be removed I think?

extension/src/engine/ws-client.ts
```diff
@@ -3,8 +3,32 @@ import { WebIdeClientJsonRpc } from './api/jsonrpc';
...
 import { IWebSocket, WebSocketMessageReader, WebSocketMessageWriter } from 'vscode-ws-jsonrpc';
+import { LogClientJsonRpc } from '@axonivy/log-view-core';
 
 export const WebSocketClientProvider = (webSocketUrl: URL) => {
+  const runtimeLogWebSocket = new WebSocket(new URL('ivy-runtime-log-lsp', webSocketUrl));
+  runtimeLogWebSocket.onopen = () => {
+    const connection = toSocketConnection(runtimeLogWebSocket);
+    const outputChannel = vscode.window.createOutputChannel('Axon Ivy Runtime Log');
+
+    function appendLog(entry: { timestamp: string; level: string; project: string; message: string; stacktrace?: string }) {
```
ivy-lmu, line 14: make a const instead of function similar to `const toSocketConnection`, and move it to the outer scope to make it more readable. nesting stuff makes it hard to read.

ivy-lmu (approved):
cool 👍 see feedback.

ivy-lmu:
> good one, see comments.
> 
> I think you can improve the output in a followup PR. See for example git output: image
> 
> there you can see some coloring stuff and there is also a log level filter. I assume you have to define a log language to achieve this, but we can research...

have a look at https://code.visualstudio.com/api/references/vscode-api#LogOutputChannel, maybe something for a followup PR

extension/src/engine/ws-client.ts
```diff
@@ -6,11 +6,13 @@ import { IWebSocket, WebSocketMessageReader, WebSocketMessageWriter } from 'vsco
 import { LogClientJsonRpc } from '@axonivy/log-view-core';
 import { RuntimeLogEntry } from '@axonivy/log-view-protocol';
 
+let outputChannel: vscode.OutputChannel;
```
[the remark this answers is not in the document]
ivy-tgr: would it be better to create already the output channel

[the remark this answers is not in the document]
ivy-lmu: good idea

PR #653, opened 2025-07-18
XIVY-16670 show project state error and provide project conversion quick fix

ivy-lmu:


ivy-lli (approved):
Cool looks good 👍

ivy-lli:
That this is error driven is fine for me, but can you disable the projects when they are already on the newest version? Or is this already the case?
I think this would help the user to understand which projects can be converted and which not.

PR #654, opened 2025-07-18
XIVY-16670 add convert project test

ivy-lmu:


playwright/tests/page-objects/editor.ts
```diff
@@ -27,4 +27,12 @@ export class Editor extends View {
     }
     await expect(this.tabLocator).toBeHidden();
   }
+
+  async activeEditorHasText(text: string) {
+    await expect(this.editorContainer()).toContainText(text);
+  }
```
ivy-lli, line 33: IMO such methods do not make much sense to have on a page object. They are just wrappers around a playwright function and sometimes not enough.
I think page objects should have method to get the locators and have functions that do multiple things. But I do not see any benefit from wrap one liner assersions :)

ivy-lli (approved):

playwright/tests/page-objects/editor.ts
ivy-lmu: good thought 👍

PR #673, opened 2025-08-04
XIVY 16480 RuntimelogOutputImprovements

ivy-tgr:
now it looks like this
image
and its possible to filter after the log levels


extension/src/engine/ws-client.ts
```diff
@@ -42,10 +42,30 @@ export const WebSocketClientProvider = (webSocketUrl: URL) => {
 };
 
 const logMessage = (entry: RuntimeLogEntry) => {
-  const logMessage = `[${entry.timestamp}][${entry.level}][${entry.project}] ${entry.message}`;
```
ivy-lmu, line 45: do you receive logs from the past? If so, the timestamp created by vscode would not match the actual one. But I'm fine with this mismatch.

ivy-lmu (approved):
nice 👍 I would also reflect this change in the test, e.g., check for error tag

PR #680, opened 2025-08-20
reduce costs to make BB happy again

ivy-lmu:


ivy-lmu:
@alexsuter seems like action usage is not the problem - fakes news. I don't merge this.

PR #686, opened 2025-08-28
Drop maven

ivy-lmu:


playwright/tests/fixtures/baseTest.ts
```diff
@@ -29,6 +29,7 @@ const runBrowserTest = async (workspace: string, take: (r: Page) => Promise<void
   await page.goto(`http://localhost:3000/?folder=${tmpWorkspace}`);
   await initialize(page);
   await take(page);
+  await page.goto('http://localhost:3000');
   await browser.close();
```
ivy-lmu, line 33: browser.close() does not close the socket connections in case of chromium.

alexsuter: not good. what do we need to do?

ivy-lmu: we need the goto

playwright/tests/fixtures/baseTest.ts
```diff
@@ -29,6 +29,7 @@ const runBrowserTest = async (workspace: string, take: (r: Page) => Promise<void
   await page.goto(`http://localhost:3000/?folder=${tmpWorkspace}`);
   await initialize(page);
   await take(page);
+  // this goto closes WebSocket connections
   await page.goto('http://localhost:3000');
```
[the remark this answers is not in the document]
ivy-lmu: no a trick.

PR #702, opened 2025-09-12
Use shared tsconfig

ivy-lli:


playwright/tests/utils/download-vscode.ts
```diff
@@ -1,4 +1,5 @@
-import { downloadAndUnzipVSCode } from '@vscode/test-electron';
-import { downloadVersion } from './download-version';
+import { downloadAndUnzipVSCode, type DownloadOptions } from '@vscode/test-electron';
 
-downloadAndUnzipVSCode(downloadVersion);
+const downloadVersion: DownloadOptions['version'] = process.env.RUN_STABLE_VERSION === 'true' ? 'stable' : 'insiders';
```
ivy-lmu, line 3: duplication

ivy-lli: Yes but, import does not work without ts-node somehow...
But IMO not a reason to have ts-node as dependency to save this one line of duplication 🙈

PR #1109, opened 2026-04-15
XIVY-18855 Set codespaces url as ivy BaseUrl

ivy-lli:
image

### DIFF
--- a/extension/src/engine/engine-runner.ts
+++ b/extension/src/engine/engine-runner.ts
@@ -1,7 +1,9 @@
 import { ChildProcess, execFile } from 'child_process';
 import Os from 'os';
 import path from 'path';
+import { Uri, env } from 'vscode';
 import { config } from '../base/configurations';
+import { stripTrailingSeparator } from '../utils/path-utils';
 import { outputChannel } from './output-channel';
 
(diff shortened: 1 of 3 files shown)


extension/src/engine/engine-runner.ts
```diff
@@ -41,13 +43,36 @@ export class EngineRunner {
...
 
+  private async resolveIvyBaseUrl() {
+    if (process.env.IVY_BASEURL) {
+      return process.env.IVY_BASEURL;
+    }
+    if (process.env.CODESPACES !== 'true' && env.appHost !== 'codespaces') {
+      return undefined;
+    }
+    try {
+      return stripTrailingSeparator((await env.asExternalUri(Uri.parse('http://localhost:8080/'))).toString());
```
ivy-lmu, line 68: localhost:8080 - does it work if the engine is running on 8081?

ivy-lmu (approved):
thx

extension/src/engine/engine-runner.ts
ivy-lli: as far as I know this is not resolved, it just needs something in and replaces all of it...

PR #1115, opened 2026-04-17
XIVY-18855 Fix custom icons blocked by csp on localhost

ivy-lli:
https://axon-ivy.atlassian.net/browse/XIVY-18855?focusedCommentId=154995

image

I think this is indeed new... maybe also since the vscode update, which maybe included a new electron version or something like that?


ivy-lmu (approved):
thx 👍

PR #1117, opened 2026-04-17
XIVY-18702 multi-step input New Project

ivy-dar:
Multi-step input dialog for New Project
- Try to use first input `projectName` for `groupId` and `projectId`. If not possible, set both to empty.
- Prohibit nested `New Project`.
If File Explorer selection is an ivy project, rightclick -> New Project will fail with warning
- Allow New Project in arbitrary folder
User can run `New Project` in folders like `.vscode`, this is not prohibited
- Move utility functions into `project-explorer/utils/util.ts`
  - `isDirectory`
  - `isSubdirectoryOrEqual`
  - `getWorkspaceFolder`
- Fix test `create-project.spec.ts`
  - Add missing namespace and id arguments in helper `addNestedProject`, since they cannot be empty inputs.

General multi-step improvements
- Custom error `MultiStepCancelledError` if users presses Esc. Implementing commands can catch it and handle it. Currently, the dialog is aborted, but not with an error (intrusive) but a log message.

### DIFF
--- a/extension/src/project-explorer/ivy-project-explorer.ts
+++ b/extension/src/project-explorer/ivy-project-explorer.ts
@@ -208,7 +208,7 @@ export class IvyProjectExplorer {
     const existingIvyProjects = await this.getIvyProjects();
     for (const existingProject of existingIvyProjects) {
       if (isSubdirectoryOrEqual(existingProject, selectedUri.fsPath)) {
-        logInformationMessage('Cannot create a new project inside an existing Axon Ivy project. Select a valid directory.');
+        logErrorMessage('Cannot create a new project inside an existing Axon Ivy project. Select a valid directory.');
         return;
       }
     }
--- a/extension/src/project-explorer/ivy-project-explorer.ts
+++ b/extension/src/project-explorer/ivy-project-explorer.ts
@@ -17,6 +17,7 @@ import { addNewProcess, type ProcessKind } from './new-process';
 import { addNewProject } from './new-project';
 import { addNewUserDialog, type DialogType } from './new-user-dialog';
 import { treeSelectionToUri, treeUriToProjectPath, type TreeSelection } from './tree-selection';
+import { getWorkspaceFolder, isDirectory, isSubdirectoryOrEqual } from './utils/util';
 
 export const VIEW_ID = 'ivyProjects';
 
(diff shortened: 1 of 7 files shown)


extension/src/project-explorer/ivy-project-explorer.ts
```diff
@@ -197,6 +198,23 @@ export class IvyProjectExplorer {
...
+    const treeSelectionUri = await treeSelectionToUri(selection);
+    const selectedUri = (await isDirectory(treeSelectionUri)) ? treeSelectionUri : await getWorkspaceFolder();
+    if (!selectedUri) {
+      logInformationMessage('No valid directory selected');
+      return;
+    }
+    const existingIvyProjects = await this.getIvyProjects();
+    for (const existingProject of existingIvyProjects) {
+      if (isSubdirectoryOrEqual(existingProject, selectedUri.fsPath)) {
+        logInformationMessage('Cannot create a new project inside an existing Axon Ivy project. Select a valid directory.');
```
ivy-lmu: would make a warning here

extension/src/project-explorer/new-project.ts
```diff
@@ -1,68 +1,93 @@
...
+  };
 
-const getWorkspaceFolder = async () => {
-  const workspaceFolders = workspace.workspaceFolders;
-  if (workspaceFolders?.length === 1 && workspaceFolders[0]) {
-    return workspaceFolders[0].uri;
-  }
-  return await window.showWorkspaceFolderPick().then(folder => folder?.uri);
-};
+  const stepGroupId: InputStep<NewProjectState> = async (input: MultiStepInput<NewProjectState>, state: NewProjectState) => {
```
ivy-lmu, line 33: I would set default value of group and artifactId to project name... sanitize project name before

playwright/tests/page-objects/explorer-view.ts
```diff
@@ -96,8 +96,8 @@ export class FileExplorer extends ExplorerView {
     await expect(newProject).toBeHidden();
 
     await this.provideUserInput(projectName);
-    await this.provideUserInput();
-    await this.provideUserInput();
+    await this.provideUserInput(projectNamespace);
```
ivy-lmu: behavior change ⚠️ as mentioned above

ivy-lmu (approved):
very nice @ivy-dar sama, looks like the multi-step framework can be reused easily. 👍 

see comment

extension/src/project-explorer/ivy-project-explorer.ts
ivy-dar: agree

extension/src/project-explorer/new-project.ts
ivy-dar: Good idea, done in
a82eb1153eb0008429d7c6e27f4af1d1278cc517
7adaf8ff1495db8ed7a6d54507156155c40d0c2b

playwright/tests/page-objects/explorer-view.ts
ivy-dar: Reverted back to original test, since `projectName` is now also used for `groupId` and `projectId`
f42c7ebef7f00ecc95aa3fcc3414a6182f5478ce

ivy-lmu:
squashed ✅  hehehe

PR #1120, opened 2026-04-20
Stabilize tests

ivy-lli:
the tests sometimes show this dialog. not sure if it is really the source that sometimes tests will fail, but I think if we can hide this it is anyways good.
image


ivy-lmu:
thx 👍

PR #1125, opened 2026-04-21
XIVY-7771 Improve process debug

ivy-lli:
- Do not set conditional breakpoints, as we currently cannot modify the condition
- Do not allow to start multiple process debug sessions
- Ask to start process debug session if breakpoint is added and none is running

### DIFF
--- a/extension/src/debug/process-debug.ts
+++ b/extension/src/debug/process-debug.ts
@@ -1,25 +1,69 @@
 import type { ExtensionContext } from 'vscode';
 import { debug, workspace } from 'vscode';
 import { registerCommand } from '../base/commands';
+import { logWarningMessage } from '../base/logging-util';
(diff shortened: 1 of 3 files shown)


extension/src/debug/process-debug.ts
```diff
@@ -1,25 +1,69 @@
...
+export function isProcessDebuggingRunning() {
+  return hasRunningProcessDebugSession;
+}
+
+export async function promptToStartProcessDebuggingIfNeeded() {
+  if (hasRunningProcessDebugSession) {
+    return false;
+  }
+
+  const selection = await window.showInformationMessage(
```
ivy-lmu: check if you can make use of logger util

ivy-lmu (approved):
👍

PR #1129, opened 2026-04-22
XIVY-18908 multi step input case map

ivy-dar:
- Add multi step input pattern for "New Case Map"
- `convertProject` change title to match design of other dialogs. No multi-step input needed


ivy-lmu (approved):
👍 wonderful

PR #1135, opened 2026-04-23
Fix tests for vscode v1.117

ivy-lli:


playwright/tests/integration/xhtml-editors.spec.ts
```diff
@@ -27,6 +27,7 @@ test('xhtml definitions', async () => {
 test('xhtml preview', async ({ page }) => {
   await editor.activateExpensiveJavaStandardMode();
   await editor.executeCommand('Axon Ivy: Deploy all Projects');
+  await editor.executeCommand('Axon Ivy: Deactivate Process Animation');
```
ivy-lmu, line 30: why? ^^

ivy-lmu (approved):
thx 👍 🥇

playwright/tests/integration/xhtml-editors.spec.ts
ivy-lli: Faster, second editor gets opened, ...

ivy-lli: IMO the animation should only be enabled on tests which tests it, otherwise it takes to much time :)

ivy-lmu: ah I did not get that a process is executed here 👍

ivy-lli: in this xhtml is a ivy component, in the preview this ivy component process is triggered 🙈

PR #1137, opened 2026-04-23
XIVY-18799 create process and casemap with spaces in namespace

ivy-dar:
Support spaces within word groups in the namespace input. Allowed for
- process
- case map

### DIFF
--- a/extension/src/project-explorer/new-case-map.ts
+++ b/extension/src/project-explorer/new-case-map.ts
@@ -11,7 +11,7 @@ import {
   type ProjectSelection,
   resolveAddCommandSelectionContext
 } from './utils/multi-step-input';
-import { validateNamespace, validateProjectArtifactName } from './utils/util';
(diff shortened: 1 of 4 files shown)


extension/src/project-explorer/new-case-map.ts
```diff
@@ -70,7 +70,7 @@ export const addNewCaseMap = async (selectionContext: AddCommandSelectionContext
       currentStep: state.currentStep,
       totalSteps: state.totalSteps,
       value: state.namespace,
-      validationFunction: validateNamespace,
+      validationFunction: validateNamespaceWithSpace,
```
ivy-lli: if you have just changed how namespaces are validated then I would not change the function name...

ivy-lli (approved):

extension/src/project-explorer/new-case-map.ts
ivy-dar: I agree, I changed it back.

PR #1144, opened 2026-04-27
XIVY-18776 create project when no workspace open

ivy-dar:
**When no workspace is open**
- Axon Ivy Projects - Treeview shows "Open Folder" option like the File Explorer Treeview

 **When a workspace is but no projects present**
- Axon Ivy Projects - Treeview shows "Create Axon Ivy Project" instead of blank
- Introduce new context key `isStarted` that complements `hasIvyProjects`
- Rename helper function `setProjectExplorerActivationCondition` to `setProjectExplorerContext`
- `IvyEngineManager.start()` is called regardless of `hasIvyProjects`. This means the engine will start even if no projects are present in the workspace. This allows for the "Create Axon Ivy Project" button without having a project already present in the workspace.
- The engine will only be started when any folder in the workspace is open
- Adjust test, now tests for a running engine in a workspace without a project

### DIFF
--- a/extension/package.json
+++ b/extension/package.json
@@ -837,6 +837,11 @@
         "view": "ivyProjects",
         "contents": "You have not yet opened a folder.\n[Open Folder](command:vscode.openFolder)",
         "when": "workbenchState == empty"
+      },
+      {
+        "view": "ivyProjects",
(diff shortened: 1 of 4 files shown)


ivy-lmu (approved):
nice

PR #1146, opened 2026-04-29
XIVY-18805 add new axon ivy project tool

ivy-lgi:
First example of an extension provided tool.


extension/src/engine/engine-manager.ts
```diff
@@ -197,7 +197,7 @@ export class IvyEngineManager {
       await this.start();
     }
     const path = newProjectParams.path;
-    this.ivyEngineApi
+    return this.ivyEngineApi
```
ivy-lgi, line 200: Needed for errors to bubble up so Copilot receives the error.

extension/src/project-explorer/new-project.test.ts
ivy-lgi, line 1: Is there a simpler way to test this? Seems like a lot of effort and fragile.

extension/src/project-explorer/new-project.ts
```diff
@@ -123,3 +133,29 @@ export const addNewProject = async (selectedUri: Uri) => {
     );
   }
 };
+
+export type NewProjectToolArgs = {
```
ivy-lgi: Not sure about the project structure and where to put this. It is strongly related to the other new project functionality which lives in `project-explorer`. But then again, it has nothing to do with the project explorer.

ivy-rew: I think we should already use this in a dedicate namespace (ai) and unit. So that it's easier to fork it into an own extension

ivy-rew (approved):
Please separate the AI related features in a common namespace -> ai/agentic

extension/src/project-explorer/new-project.test.ts
ivy-lmu: questionable if you want to test that here. in the end, you just call the createProject API, which is already tested. And you’re planning an e2e tool test...

extension/src/project-explorer/new-project.ts
ivy-lmu: yep I would isolate all the tool code.

ivy-lmu (approved):

ivy-dar (commented):
Looks good to me 👍

PR #1147, opened 2026-04-29
XIVY-18767 Dialog preview qg

ivy-lli:
- Adjust preview icon
- Also show command and icon on f.json
- Do not show command and icon on components (when include a <cc:interface


ivy-lmu (approved):
👍

PR #1152, opened 2026-04-30
XIVY-18937 set default project namespace after project selection

ivy-dar:
Set the `namespace` input default to the namespace of the selected project, if the user did not select a project via tree selection (right click)
- Implemented for all commands that have a project selection and a namespace input
  - New Business / Callable / Webservice Process
  - New Data Class / Entity Class
  - New Html / Form / Offliine Dialog
  - New Case Map
- Initially only takes effect if the user has not selected a valid filepath via rightclick in File Explorer or Axon Ivy Projects tree. If this selection exists, it is respected over the default project namespace.
- Updates the proposed `namespace` whenever the project selection in the earlier step changes.
- As long as the project selection does not change, the user entered namespace will be conserved.

### DIFF
--- a/playwright/tests/page-objects/explorer-view.ts
+++ b/playwright/tests/page-objects/explorer-view.ts
@@ -100,11 +100,17 @@ export class FileExplorer extends ExplorerView {
     await this.provideUserInput();
   }
 
-  async addProcess(processName: string, kind: 'Business Process' | 'Callable Sub Process' | 'Web Service Process', namespace?: string) {
+  async addProcess(
+    processName: string,
+    kind: 'Business Process' | 'Callable Sub Process' | 'Web Service Process',
+    defaultNamespaceExpected: string = 'prebuiltProject',
+    namespace?: string
+  ) {
     await this.selectNode('config');
     await this.executeCommand('Axon Ivy: New ' + kind);
     await this.selectNthVisibleItemFromQuickPick(0);
     await this.provideUserInput(processName);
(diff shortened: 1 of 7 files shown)


playwright/tests/page-objects/explorer-view.ts
```diff
@@ -100,11 +100,20 @@ export class FileExplorer extends ExplorerView {
...
-  async addProcess(processName: string, kind: 'Business Process' | 'Callable Sub Process' | 'Web Service Process', namespace?: string) {
+  async addProcess(
+    processName: string,
+    kind: 'Business Process' | 'Callable Sub Process' | 'Web Service Process',
+    namespace?: string,
+    defaultNamespaceExpected?: string
+  ) {
+    if (defaultNamespaceExpected === undefined) {
+      defaultNamespaceExpected = 'prebuiltProject';
+    }
```
ivy-dar, line 111: Not pretty, but the other way would be to supply an options element like this
```addProcess(options: {
  processName: string
  kind: 'Business Process' | 'Callable Sub Process' | 'Web Service Process',
  defaultNamespaceExpected: string = 'prebuiltProject'
  namespace?: string,
})
```
in order to call addProcess like this
```typescript
await explorer.addProcess({processName, kind:'Business Process', namespace:'parent1/parent2'});
```
which is even uglier and intrusive...

ivy-lmu (approved):
desk review ;) 👍

PR #1163, opened 2026-05-07
Feature/implement install market product

ivy-dar:
- Implement multi step input for "Install Market Product"
  - The dialog can be executed in an empty workspace (no ivy project)
  - If the user selects a maven-dependency product that requires an existing ivy project, the dialog aborts and informs the user to create an ivy project first.
- Add `canSelectMany` support in `multi-step-input.showQuickPick`
  - Remove property `activeItem`, replaced by `activeItems`
  - Adjust usages of `showQuickPick` to account for changes
- Fix typo in orval.config.ts, regenerate schemas `market-client.ts` and `client.ts`
- Adjust tests for installing a market product
- Make tests less flaky by removing `selectNthVisibleItemFromQuickPick` and replace in all ocurrences with hardcoded values

## Work remaining
- [x] Add playwright test for both maven-import only and maven-dependency case
- [x] Activate `isIvyProjectSelectionRequired` check to skip project step if applicable

## Open discussions regarding Market
See [XIVY-18938](https://axon-ivy.atlassian.net/browse/XIVY-18938)

[XIVY-18938]: https://axon-ivy.atlassian.net/browse/XIVY-18938?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


extension/src/market/import-market.ts
```diff
@@ -101,6 +89,177 @@ async function replaceDynamicVersion(productJson: string) {
...
+  };
+
+  const stepDependentProject: InputStep<InstallMarketProductState> = async (
+    input: MultiStepInput<InstallMarketProductState>,
+    state: InstallMarketProductState
+  ) => {
+    // Note: Currently, engine does not support missing ivy project dependency
+    // if (!isIvyProjectSelectionRequired(state.projects ?? [])) {
+    //   return;
+    // }
```
ivy-dar: We can maybe change the behaviour of the backend to accept a missing project dependency if only maven-imports are selected. If this is true, this part would work as expected and only prompt the user for a project if at least one maven-dependency is selected.

playwright/tests/page-objects/page-object.ts
```diff
@@ -67,6 +67,10 @@ export class PageObject {
     return this.page.locator('div.quick-input-list');
   }
 
+  quickInputHeader(): Locator {
```
ivy-lmu: I wouldn't add a method if the thing is only used once in your test

playwright/tests/integration/install-product.spec.ts
```diff
@@ -16,7 +16,16 @@ test.describe('Market Product installation', () => {
     await explorer.selectNode('resources');
     await explorer.installProduct('connectivity-demo');
     await explorer.provideUserInput('14.0.0-SNAPSHOT');
-    await explorer.provideUserInput('connectivity-demos$');
+    const header = explorer.quickInputHeader();
+    const checkbox = header.getByRole('checkbox', { name: 'Toggle all checkboxes' });
+    await expect(checkbox).toBeVisible();
```
ivy-lmu, line 21: I guess you can use `await expect(checkbox).toBeChecked();`

playwright/tests/integration/install-product.spec.ts
```diff
@@ -16,7 +16,16 @@ test.describe('Market Product installation', () => {
...
+    const ariaChecked = await checkbox.getAttribute('aria-checked');
+    if (ariaChecked == 'false') {
+      await checkbox.click();
+    }
```
ivy-lmu, line 25: don't you hit check all above? or is this something different?

ivy-lmu (approved):
good 👍 market details to be discussed with @ivy-rew

playwright/tests/page-objects/page-object.ts
ivy-dar: Very true, removed

playwright/tests/integration/install-product.spec.ts
ivy-dar: The "Toggle all" Checkbox is not automatically checked. It depends on the market product. I thought a safer way to make sure its toggled is to find it, and toggle if not yet toggled.

extension/src/market/import-market.ts
```diff
@@ -1,9 +1,47 @@
...
+interface InstallMarketProductState extends MSStateBase {
+  product?: ProductSelection;
+  productJson?: string;
+  version?: string;
+  projects?: ProductProjectSelection[];
+  projectsSearchString?: string;
+  changedProjectSelection?: boolean;
+  dependentProject?: ProjectSelection;
+  dependentProjectFilterText?: string;
+  // productJson?: string;
```
ivy-lmu: remove?

ivy-lmu (approved):
👍

PR #1167, opened 2026-05-12
Install Local Market Product - apply multi step dialog

ivy-dar:
Implement the multi step dialog pattern for the "Install Local Market Product" command

### DIFF
--- a/extension/src/project-explorer/ivy-project-tree-data-provider.ts
+++ b/extension/src/project-explorer/ivy-project-tree-data-provider.ts
@@ -63,10 +63,13 @@ export class IvyProjectTreeDataProvider implements TreeDataProvider<Entry> {
   }
 
   private async findIvyProjects(): Promise<string[]> {
-    return (await workspace.findFiles(IVY_RPOJECT_FILE_PATTERN, this.excludePattern, this.maxResults))
-      .filter(p => isIvyProject(p))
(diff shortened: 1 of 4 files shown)


extension/src/project-explorer/ivy-project-tree-data-provider.ts
```diff
@@ -63,10 +63,13 @@ export class IvyProjectTreeDataProvider implements TreeDataProvider<Entry> {
...
-      .filter(p => isIvyProject(p))
-      .map(p => path.dirname(p.fsPath))
-      .sort();
+    return Array.from(
+      new Set(
+        (await workspace.findFiles(IVY_RPOJECT_FILE_PATTERN, this.excludePattern, this.maxResults))
+          .filter(p => isIvyProject(p))
+          .map(p => path.dirname(p.fsPath))
+      )
+    ).sort();
```
ivy-dar, line 72: When importing a product mutliple times, it happend that both an .ivyproject and .project file were present. Without the dedpulication, the dialog showed the same project (same name, same fs path) twice.

ivy-lmu: probably we will get other problems on engine side if both files are present. but ok 👍

ivy-lmu (approved):

PR #1169, opened 2026-05-13
XIVY-18967 use engine version to retreive available products

ivy-dar:
- Regenreate engine and market client
- Use web-ide engine version endpoint to retrieve current engine version
- Use engine version as the version used in install market product. desigern version is fallback
- Use this version to fetch only available versions from market


orval.config.ts
```diff
@@ -11,7 +11,7 @@ export default defineConfig({
   ivyOpenApi: {
     input: {
       target: 'target/engine/openapi.json',
-      filters: { tags: ['web-ide'] }
+      filters: { tags: ['web-ide', 'engine'] }
```
ivy-lmu: filter for path not possible?

ivy-lmu (approved):
👍

orval.config.ts
ivy-dar: Not directly. Should we add the tag "web-ide" to the /engine/info method?

ivy-lmu: if you can add multiple tags, yes please

PR #1178, opened 2026-05-21
add new process tool

ivy-lgi:
Also, return the path of the resource created by the tools so Copilot knows immediately where it is, instead of having to search for it.


ivy-lmu (approved):
🤖 👍

ivy-dar (approved):

PR #1179, opened 2026-05-21
do not run rest generation if spec is not provided

ivy-lmu:


extension/src/editors/restclient-editor/webview-communication.ts
```diff
@@ -69,9 +69,12 @@ class RestClientWebSocketForwarder extends WebSocketForwarder {
   }
 }
 
-async function generateClient(payRaw: string, document: TextDocument) {
+const generateClient = async (payRaw: string, document: TextDocument) => {
```
ivy-lmu, line 72: or should it be fixed on client side? `generateOpenApiClient` is always fired, no matter if spec is provided.

ivy-rew: yep, I think it would be better to not fire this event no openapi is here.

ivy-dar (approved):
Will be fixed in the restclient-editor

ivy-lmu:
https://github.com/axonivy/restclient-editor/pull/83

PR #1182, opened 2026-05-21
add market skill

ivy-lgi:


ivy-lgi:
First extension provided skill :tada: 

I think for the market it makes sense to have a skill that concisely explains what it's about, instead of providing tools. Copilot is responsible to fetch the API spec and use its endpoints to achieve its goal, so the skill is very simple and agnostic.

I'm sure some more details like how to determine the current designer version might be helpful down the line but this works for now.

ivy-rew:
Please be a role-model for future contributors; supply a test with a prompt that is brightly handled -> e.g. "install a jira-connector and fetch pending issues".

ivy-lgi:
> Please be a role-model for future contributors; supply a test with a prompt that is brightly handled -> e.g. "install a jira-connector and fetch pending issues".

As I understood from the QG, I am supposed to just quickly implement some tools and skills without worrying about testing. We don't have any test infrastructure here for this yet. If I start with that now, I don't think I will be able to finish it until the next Sprint starts.

ivy-rew:
> > Please be a role-model for future contributors; supply a test with a prompt that is brightly handled -> e.g. "install a jira-connector and fetch pending issues".
> 
> As I understood from the QG, I am supposed to just quickly implement some tools and skills without worrying about testing. We don't have any test infrastructure here for this yet. If I start with that now, I don't think I will be able to finish it until the next Sprint starts.

Yes this was one voice, but we didn't align on it. But we already had stories planned and prioritized on testing which I think is really more important. As soon as we have an answer on testing; any team can support us here.

However; I don't mind if you pack this as example by now as you already have it.

ivy-lgi:
I close this PR and stale the branch so it can be used as a starting point in the future because:
1. It seems like we are not aligned on what to do.
2. @ivy-dar noticed that Copilot cannot handle products that need to be installed as `maven-import`. Multiple ways to fix this are possible but it also needs alignment.
3. I am no longer planned to work on this and will move to other topics with the start of the next sprint.

ivy-dar:
> I close this PR and stale the branch so it can be used as a starting point in the future because:
> 
> 1. It seems like we are not aligned on what to do.
> 2. @ivy-dar noticed that Copilot cannot handle products that need to be installed as `maven-import`. Multiple ways to fix this are possible but it also needs alignment.
> 3. I am no longer planned to work on this and will move to other topics with the start of the next sprint.

Regarding point 2: I think before moving on with an AI integration, it should be clearer how maven-import vs maven-dependencies are handled in the future

PR #1200, opened 2026-06-03
parallel browser tests

ivy-lmu:


playwright/tests/integration/create-process.spec.ts
```diff
@@ -14,7 +14,7 @@ test.describe('Create Process', () => {
     processEditor = new ProcessEditor(page, `${processName}.p.json`);
   });
 
-  test('Add business process, execute, edit and redeploy', async () => {
+  test('Add business process, execute, edit and redeploy', { tag: '@serial' }, async () => {
```
ivy-lmu, line 17: simulating process models, i.e. start processes, seem to be flaky. so let's run it serial.

PR #1213, opened 2026-06-10
XIVY-18535 handle service.info

ivy-rew:
adapt: https://github.com/axonivy/webservice-codegen/pull/26


ivy-lli (approved):
Ah vscode adds this param :D question answerd ;)

PR #1217, opened 2026-06-11
XIVY-18535 pick URI: windows compatible 

ivy-rew:
Bruno found this; we can't deal properly with windows file paths ... so we rather create an URI.

works:
vsc-codegen-win

did not work:
not-work-ws

### DIFF
--- a/extension/src/editors/file-picker.ts
+++ b/extension/src/editors/file-picker.ts
@@ -31,5 +31,5 @@ export async function pickFile(request: FilePickRequestLike, document: TextDocum
     return relativePath;
   }
 
-  return selected.fsPath;
+  return selected.toString();
 }


ivy-lli (approved):

ivy-dar (approved):

extension/src/editors/file-picker.ts
```diff
@@ -31,5 +31,5 @@ export async function pickFile(request: FilePickRequestLike, document: TextDocum
     return relativePath;
   }
 
-  return selected.fsPath;
```
ivy-lmu, line 34: usually, fsPath is a good choice, why not here?

ivy-rew: the string is passed to the wsdl meta endpoint, which seems to be unhappy with this windows path. Maybe it should be double escaped? At any rate we support anyway URIs or Paths. So switching to URI shouldn't hurt. 

---

Basically I'd argue the URI is always the better format to pass around. As its more rigid than just a string. Here we forward the thing to meta-endpoint and also to vsc-integration for generation -> then to plugin. The farther we pass, the string is very limited in its expressiveness.

PR #1231, opened 2026-06-18
XIVY-18804 relax java dependency

ivy-lmu:


extension/package.json
```diff
@@ -72,7 +72,7 @@
   "activationEvents": [
     "onStartupFinished"
   ],
-  "extensionDependencies": [
+  "extensionPack": [
```
ivy-lmu, line 75: the listed extensions will be installed when our extension is installed. However, activation of our extension no longer waits for those extensions to be activated.

- you cannot disable a listed extensionPack extension without also disabling our extension (at least through the UI).
- a listed extension can still be uninstalled, which is why I added a programmatic check to handle that case (only for java server).

ivy-lli (approved):
thx 👍

ivy-dar (approved):
Nice one 👍

PR #1235, opened 2026-06-18
fix devcontainer external url

ivy-lmu:
port mapping is dynamic, so let's evaluate the external engine url.
image


ivy-lli (approved):
Ok, but how do you now evaluate the devcontainter external url?

ivy-lmu:
> Ok, but how do you now evaluate the devcontainter external url?

If a devcontainer maps port 8080 to 1234, then asExternalUri(localhost:8080) resolves to localhost:1234.
Likewise in codespaces, asExternalUri returns the correct external URI for the local port you pass in.

PR #1242, opened 2026-06-24
XIVY-19115 MCP wrapper serve: vsc.llm.tools :globe_with_meridians: 

ivy-rew:
Bring VSC llm.tools to live in third-party harnesses:
- MCP is an opt-in enablement ... nothing runs by default

copilot-mcp-claude


extension/src/index.ts
```diff
@@ -48,6 +50,9 @@ export async function activate(context: ExtensionContext): Promise<MessengerDiag
     registerCommand('ivy.showStatusBarQuickPick', context, (visibleOptions?: string[]) => StatusBar.showStatusBarQuickPick(visibleOptions));
 
     registerTools(context);
+    const localMcp = config.localMcp();
+    localMcpServer = new LocalMcpServer();
+    await localMcpServer.start(localMcp);
```
ivy-lmu: really need to await?

ivy-lmu (approved):
👍

ivy-lmu:
@ivy-cst fyi

extension/src/index.ts
ivy-rew: no you are right, let's try to start ... but this infrastructure is not critical. having a separate toast for it on failure should be sufficient.

extension/src/ai/tools/local-mcp.ts
```diff
@@ -0,0 +1,269 @@
...
+    }
+    return { type: 'object', properties: {} };
+  }
+
+  private async invokeOwnToolHeadless(name: OwnToolName, args: unknown): Promise<{ content: unknown[] }> {
+    if (!args || typeof args !== 'object' || Array.isArray(args)) {
+      throw new Error(`Tool '${name}' expects arguments as an object`);
+    }
+
+    switch (name) {
```
ivy-lgi: Maybe introduce some kind of data structure that holds the tools instead of this big switch statement. Could be used something like this: `tools.find(tool => tool.name === name).invoke();`

doc/mcp/mcp.md
```diff
@@ -0,0 +1,45 @@
...
+
+The local MCP is not enabled by default; if you like to access
+the AI tools, enable it by setting the `axonivy.localMcp.enabled: true`
+
+```json
+{
+  "axonivy.localMcp.enabled": true,
+  "axonivy.localMcp.host": "127.0.0.1",
+  "axonivy.localMcp.port": 32140,
+  "axonivy.localMcp.exposeAllTools": false
```
ivy-lgi, line 17: Do we really want to expose all tools? If this is something Microsoft wants, I assume they would have implemented this themselves. I suspect Microsoft might not like this (or they will add it in the future).

I would just provide our tools and not even go through the VS Code `lm` stack if it can be avoided. I don't think I like meddling with that.

Also, calling other tools non-headless could result in interactive confirmation popups. At the very least this should be documented.

extension/src/ai/tools/tools.ts
```diff
@@ -3,16 +3,29 @@ import { NewDataClassTool } from './new-data-class';
 import { NewDialogTool } from './new-dialog';
 import { NewProcessTool } from './new-process';
 import { NewProjectTool } from './new-project';
+import { isOwnToolName, OWN_TOOL_NAMES, type OwnToolName } from './tool-ids';
 
 export function registerTools(context: ExtensionContext) {
```
ivy-lgi, line 60: A data structure could handle this as well.

doc/mcp/.mcp.json
ivy-lgi, line 1: Why have the MCP server configured for this repo?

README.md
```diff
@@ -32,6 +32,11 @@ Then simply start the `Run Extension` launch config, which will also activate th
 
 If you do not want to use the embedded engine as backend, you can define an alternative under `VS Code Settings / Axon Ivy`. Simply uncheck `Engine: Run By Extension` and set `Engine: Url` you want to use.
 
+## MCP setup for Copilot CLI
+
+You can expose the extension tools through MCP and connect Copilot CLI to that local endpoint.
```
ivy-lgi: Maybe mention not just Copilot CLI but that it works for any harness supporting tools. I think that it can be used with any harness is the big selling point.

ivy-lgi (commented):
Very cool 🔥 But I'm hesitant about exposing all tools of VS Code. Maybe we should align on that. And some other suggestions.

doc/mcp/mcp.md
ivy-rew: On the long term you are totally right. Most tools will not be useful or ready to use.
That's why the default is false: so only our well known tools are exposed. 

We just have the switch here to play with the other tools if necessary... no strong feelings to keep it for long time. But I think it would help to get experience with them whether there are any problems or if there are at least a selection of standard tools that could be handy.

doc/mcp/.mcp.json
ivy-rew: you are right, It's missleading like this ... I'll move it to the docs directory where it is mentioned. It should just serve as example.

extension/src/ai/tools/local-mcp.ts
ivy-rew: yes, moving the thing out of the MCP makes it much easier to maintain

doc/mcp/mcp.md
ivy-lgi: Fine with me for now :+1:

ivy-lgi (approved):
Nice, thanks 👍

ivy-rew:
fyi @nhthinh-axonivy ; MCP is in and ready to be tested with real-world prompts

nhthinh-axonivy:
> fyi @nhthinh-axonivy ; MCP is in and ready to be tested with real-world prompts

Lightning work! I'm working on review all the skill to categories what should be standard or belong to other group e.g: customer service...

PR #1253, opened 2026-07-01
XIVY-18991 Install Market Product trigger Java extension to build after import

ivy-dar:
- Related to: https://github.com/axonivy/core/pull/13545
- If empty workspace and then import -> Java extension does not start -> trigger manually by activating it if not yet activated and trigger reload of projects in java extension


ivy-cst (approved):
Looks good, but @ivy-lmu should have a look.

ivy-lli (approved):
Looks good 👍

extension/src/engine/engine-manager.ts
```diff
@@ -202,6 +202,8 @@ export class IvyEngineManager {
       { text: 'Importing market product' },
       async () => await this.ivyEngineApi?.installMarketProduct(input)
     );
+    await this.importJavaProjects();
+    await executeCommand('ivyProjects.refreshEntry');
```
ivy-lmu: do you really need to execute a command here? try to make a direct call

ivy-lmu (approved):
👍 thanks

extension/src/engine/engine-manager.ts
ivy-dar: Direct call where? In `IvyProjectExplorer.refresh()`?

It is a private method, only exposed through the command, therefore I thought the least intrusive solution is to simply call the command

ivy-lmu (approved):
gracias 👍

PR #1258, opened 2026-07-06
test: add MCP integration test

nhthinh-axonivy:
## What this PR does

This PR adds an **extension-host integration test suite** (`extension/src/test-integration`) to verify that a **real AI coding agent** can discover and invoke our MCP tools from natural-language prompts.

Unlike the existing Vitest suite, which validates the MCP protocol using mocks, these tests launch a real VS Code instance via `@vscode/test-electron`, start the embedded Axon Ivy Engine, and run the GitHub Copilot CLI against it.

The current suite verifies that the following tools are correctly discovered and executed:

* `new_axon_ivy_project`
* `new_data_class`
* `new_process`
* `new_dialog`

The generated project files are used to verify successful execution.

## Why Mocha?

The tests run inside a real VS Code extension host using `@vscode/test-electron`. Mocha is the standard runner for this environment and is the approach used by Microsoft's VS Code extension samples. Vitest is still used for unit tests.

## How to run

```bash
cd extension

# Full suite
GH_TOKEN=$(gh auth token) pnpm run test:integration

# Without GH_TOKEN, AI tests are skipped
pnpm run test:integration
```

**Prerequisites:**

* GitHub Copilot access (only for the AI-driven tests)

### DIFF
--- a/extension/src/test-integration/runTest.ts
+++ b/extension/src/test-integration/runTest.ts
@@ -1,0 +1,57 @@
+import { downloadAndUnzipVSCode, runTests } from '@vscode/test-electron';
+import fs from 'fs';
+import os from 'os';
+import path from 'path';
+
+async function main() {
+  const javaHome = process.env.IVY_JAVA_HOME ?? process.env.JAVA_HOME;
+  if (!javaHome) {
+    throw new Error('JAVA_HOME (or IVY_JAVA_HOME) must be set to a Java 25 installation to run this test.');
+  }
+
+  // process.cwd() is the extension package root (this script is run via a package script from `extension/`).
+  const extensionDevelopmentPath = process.cwd();
+  const extensionTestsPath = path.resolve(process.cwd(), 'dist-test-integration/suite/index.js');
+  const fixtureWorkspacePath = path.resolve(process.cwd(), 'src/test-integration/fixtures/mcpEnabled');
+  const workspacePath = await prepareWorkspace(fixtureWorkspacePath, javaHome);
+
+  const vscodeExecutablePath = await downloadAndUnzipVSCode();
+
+  await runTests({
+    vscodeExecutablePath,
+    extensionDevelopmentPath,
+    extensionTestsPath,
+    launchArgs: [
+      workspacePath,
+      '--disable-extensions',
+      '--disable-workspace-trust',
+      '--disable-gpu',
+      '--skip-welcome',
+      '--skip-release-notes'
+    ],
+    extensionTestsEnv: {
+      IVY_JAVA_HOME: javaHome,
+      // process.cwd() inside the spawned extension host is not guaranteed to be the extension package root.
+      TEST_SUITE_DIR: path.resolve(process.cwd(), 'dist-test-integration/suite'),
+      // Optional: only present when running the exploratory Copilot CLI test (see local-mcp-copilot.test.ts).
+      ...(process.env.GH_TOKEN ? { GH_TOKEN: process.env.GH_TOKEN } : {})
+    }
+  });
+}
+
+async function prepareWorkspace(fixtureWorkspacePath: string, javaHome: string): Promise<string> {
+  const tmpWorkspace = await fs.promises.mkdtemp(path.join(os.tmpdir(), 'mcp-extension-host-workspace-'));
+  await fs.promises.cp(fixtureWorkspacePath, tmpWorkspace, { recursive: true });
+
+  const settingsPath = path.join(tmpWorkspace, '.vscode/settings.json');
+  const settings = JSON.parse(await fs.promises.readFile(settingsPath, 'utf8'));
+  settings['java.jdt.ls.java.home'] = javaHome;
+  await fs.promises.writeFile(settingsPath, JSON.stringify(settings, null, 2));
+
+  return tmpWorkspace;
+}
+
+main().catch(error => {
+  console.error('Extension-host integration test failed to run:', error);
+  process.exit(1);
+});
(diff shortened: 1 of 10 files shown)


nhthinh-axonivy:
Hi @ivy-rew, @ivy-lgi I have draft some idea how we can do the integration test for AI coding agent to use our mcp tools.

extension/src/test-integration/suite/local-mcp-copilot.test.ts
```diff
@@ -0,0 +1,141 @@
...
+ */
+suite('Local MCP Server — AI-driven tool pickup (exploratory)', () => {
+  let projectsDir: string;
+  let projectPath: string;
+
+  suiteSetup(async function () {
+    if (!process.env.GH_TOKEN) {
+      this.skip();
+      return;
+    }
```
ivy-lgi, line 32: I'd fail hard if the token is not set. This test suite needs to be triggered explicitly anyway and if for some reason the token goes missing inside the build infrastructure these tests will no longer be running without notice.

extension/src/test-integration/suite/local-mcp-copilot.test.ts
```diff
@@ -0,0 +1,141 @@
...
+  });
+
+  suiteTeardown(async () => {
+    if (projectsDir) {
+      await fs.promises.rm(projectsDir, { recursive: true, force: true });
+    }
+  });
+
+  test('picks up new_axon_ivy_project from a prompt', async function () {
+    this.timeout(120_000);
```
ivy-lgi, line 51: Can this be set globally instead of having to write it for every test?

extension/src/test-integration/runTest.ts
```diff
@@ -0,0 +1,57 @@
...
+      '--disable-gpu',
+      '--skip-welcome',
+      '--skip-release-notes'
+    ],
+    extensionTestsEnv: {
+      IVY_JAVA_HOME: javaHome,
+      // process.cwd() inside the spawned extension host is not guaranteed to be the extension package root.
+      TEST_SUITE_DIR: path.resolve(process.cwd(), 'dist-test-integration/suite'),
+      // Optional: only present when running the exploratory Copilot CLI test (see local-mcp-copilot.test.ts).
+      ...(process.env.GH_TOKEN ? { GH_TOKEN: process.env.GH_TOKEN } : {})
```
ivy-lgi, line 37: What kind of token is this? Will this use my personal Copilot subscription? I think we should change it to use a specific API key instead.

ivy-lgi: I think in the end we want to run this in some kind of sandbox, a docker container probably. In it, Install GitHub copilot, configure an API key for Copilot, and run tests within. Both for local development/experiments and when we create a Jenkins build job to run these.

Seems much to easy to accidentally use my own tokens currently and to create new tools/experiment, we also need a dedicated provider instead of using our personal credits.

extension/src/test-integration/suite/local-mcp-copilot.test.ts
```diff
@@ -0,0 +1,141 @@
...
+    await runCopilotPrompt(
+      `Create an Axon Ivy project named copilot-demo at ${projectsDir}, group id com.test, project id com.test.copilot`,
+      projectsDir
+    );
+
+    projectPath = path.join(projectsDir, 'copilot-demo');
+    assert.ok(fs.existsSync(path.join(projectPath, 'pom.xml')), 'expected Copilot to create the project via the MCP tool');
+  });
+
+  test('picks up new_axon_ivy_data_class from a prompt', async function () {
```
ivy-lgi, line 61: Does this (and following) test(s) use the project created by the previous test?
If so, this introduces dependencies between tests and I would instead create an empty test project to be opened and used instead (apart from the `new_axon_ivy_project` test maybe)

extension/src/test-integration/suite/local-mcp-copilot.test.ts
```diff
@@ -0,0 +1,141 @@
...
+    this.timeout(120_000);
+    const before = await countMatches(projectPath, '**/*.d.json');
+
+    await runCopilotPrompt(`In the Axon Ivy project at ${projectPath}, create a data class named Customer in namespace demo.data`, projectPath);
+
+    const after = await countMatches(projectPath, '**/*.d.json');
+    assert.ok(after > before, 'expected Copilot to create a new .d.json data class file via the MCP tool');
```
ivy-lgi, line 68: Instead of only asserting there are more, assert the actual name and content (same for the other tests).

ivy-lgi (commented):
Good start! 👍 
Some questions and suggestions.

And i don't know very much about these configs, build scripts, etc. Maybe @ivy-lmu can have a quick look?

ivy-lgi:
Also, I think the whole testing stuff should be moved into a new package in `root`. Similar to how `playwright` does it and be called something like `ai-integration-test`.

ivy-lmu:
> Also, I think the whole testing stuff should be moved into a new package in `root`. Similar to how `playwright` does it and be called something like `ai-integration-test`.

I would also isolate the integration tests, as recommended by @ivy-lgi. How are you planning to run test:integration in CI?

ivy-lmu:
if no work is planned, please stale this branch

ivy-lgi:
@nhthinh-axonivy I prefixed the branch with `stale/` so tests are no longer running every night. The branch still exists under `stale/feat/add-mcp-test`.

nhthinh-axonivy:
> @nhthinh-axonivy I prefixed the branch with `stale/` so tests are no longer running every night. The branch still exists under `stale/feat/add-mcp-test`.

Yes, sorry, too much going on now so I completely forgot this.

PR #1260, opened 2026-07-06
Rewrite page objects

ivy-lli:


playwright/tests/integration/market/install-product.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+import { ProcessEditor } from '~/page-objects/process-editor';
+
+test('Install product without maven-dependency from Market website', async ({ wsPage }) => {
+  const explorer = new FileExplorer(wsPage);
+  await explorer.selectNode('resources');
+  await explorer.installProduct('connectivity-demo');
+  await wsPage.provideUserInput('14.0.0-SNAPSHOT');
+  const header = wsPage.page.locator('div.quick-input-header');
+  const checkbox = header.getByRole('checkbox', { name: 'Toggle all checkboxes' });
+  await checkbox.check();
```
ivy-lli, line 12: any reason why this was:
```ts
await expect(checkbox).toBeVisible();
const ariaChecked = await checkbox.getAttribute('aria-checked');
if (ariaChecked == 'false') {
  await checkbox.click();
}
await expect(checkbox).toHaveAttribute('aria-checked', 'true');
```

ivy-lmu: I guess it was flaky, but maybe @ivy-dar can tell

ivy-lmu (approved):
cool thx 👍

playwright/tests/integration/market/install-product.spec.ts
ivy-dar: Depending on the product you install, the checkbox was sometimes already checked (e.g. all items were preselected), sometimes not (I think it depends on the specification of the product you install).
I wanted to make sure that the click does not unselect an already checked box, but with the check() method I think this should not happen.

PR #1267, opened 2026-07-10
fix i18n version

ivy-lmu:


package.json
```diff
@@ -48,12 +48,12 @@
     "@vitejs/plugin-react": "6.0.3",
     "adm-zip": "0.5.18",
     "globals": "17.7.0",
-    "i18next": "^26.3.4",
```
ivy-lmu, line 51: with that, two versions of i18next (26.3.4 and 26.3.5) became available in CI, leading to a conflict.

ivy-lli (approved):
👍

PR #1272, opened 2026-07-13
XIVY-18688 no more compile after unpack

ivy-dar:
- Related to PR https://github.com/axonivy/core/pull/13670
- `compile` is removed from install market product, no need for `doCompile` flag
- Regenrate clients
- Remove all usages of `doCompile`


ivy-lmu (commented):
thx 👍

PR #1273, opened 2026-07-14
XIVY-19207 VSCode import IAR and ZIP

ivy-dar:
- Add `importIvyProject` command
- Add `importIvyProject` function in `ivy-project-explorer` for .iar and .zip files
  - Only allows import into root workspace, no nested imports
  - Does currently not properly support multi-root vsc workspaces. If you open vsc workspace "A" and then add "B" as a vsc workspace and try to import an ivy project into "B", you will get a `Workspace does not exist` error. Import into "A" will work.
- Add `download-iar.ts` utility to download an .iar file for the test of `importIvyProject`
- Move pnpm script `test:playwright:download:vscode` into a playwright setup in `playwrigth/tests/test.setup.ts` following the [playwright docs](https://playwright.dev/docs/test-global-setup-teardown#option-1-project-dependencies)
- Move env vars into `Jenkinsfile` and keep `package.json` clean


extension/src/project-explorer/import-process.ts
```diff
@@ -10,18 +10,21 @@ export const importNewProcess = async (projectDir: string) => {
...
-const collectImportBpmnProcessParams = async (projectDir: string): Promise<ImportProcessBody> => {
+const collectImportBpmnProcessParams = async (projectDir: string): Promise<ImportProcessBody | undefined> => {
   const bpmnXmlFile = await window.showOpenDialog({
     canSelectMany: false,
-    openLabel: 'Select BPMN XML File to Import'
+    title: 'Select BPMN .bpmn file to import',
+    openLabel: 'Import BPMN Process',
+    filters: {
+      'BPMN Files': ['bpmn']
+    }
```
ivy-dar, line 20: Is this correct? Are BPMN files always .bpmn or are other file extensions allowed?

package.json
```diff
@@ -27,8 +27,7 @@
     "test:ci": "pnpm run -r test:ci",
     "test:playwright": "pnpm run --filter @axonivy/vscode-designer-playwright test:ui --project=integration-serial --project=integration-parallel",
     "test:playwright:browser": "RUN_IN_BROWSER=true pnpm run --filter @axonivy/vscode-designer-playwright test:ui",
-    "test:playwright:download:vscode": "pnpm run --filter @axonivy/vscode-designer-playwright download:vscode",
-    "test:playwright:performance": "pnpm run --filter @axonivy/vscode-designer-playwright test:ui --project=performance",
+    "test:playwright:performance": "set SKIP_IAR_DOWNLOAD=true && pnpm run --filter @axonivy/vscode-designer-playwright test:ui --project=performance",
```
ivy-dar: This is a windows specific solution, but I suppose we'll never run the performance test outside of a windows build.

playwright/tests/utils/download-iar.ts
```diff
@@ -0,0 +1,67 @@
...
+    const patterns = rawPattern
+      .split(',')
+      .map(s => s.trim())
+      .filter(Boolean);
+    await downloadIar(url, patterns, console.log);
+  } catch (error) {
+    console.error('Failed to download IAR');
+    console.error('URL: ' + url);
+    console.error('Filename patterns: ' + rawPattern);
+    console.error('Error:', error);
```
ivy-dar: On purpose continue if iar download fails. Log and let the tests fail that depend on the iar file.

extension/src/project-explorer/import-process.ts
ivy-lmu: no idea. @ivy-rew what do you mean?

playwright/tests/page-objects/explorer-view.ts
```diff
@@ -38,6 +38,11 @@ abstract class ExplorerView {
     await expect(node).toBeVisible();
   }
 
+  async hasNodeExact(name: string) {
```
ivy-lmu, line 36: maybe this one can be used anywhere?

playwright/tests/page-objects/workspace-page.ts
```diff
@@ -49,6 +49,10 @@ export class WorkspacePage {
     return this.page.locator('div.notification-toast-container');
   }
 
+  async hasNoLoggedErrors() {
+    await expect(this.toasts).toBeHidden();
```
ivy-lmu: I generally avoid adding single use methods to page objects

playwright/tests/utils/download-iar.ts
```diff
@@ -0,0 +1,67 @@
+import fs from 'fs';
+import JSZip from 'jszip';
+import path from 'path';
+import { emptyDownloadIarFilenameUpToDate as targetIarFilename, empty as targetIarPath } from '../workspaces/workspace';
+
+const downloadIar = async (
```
ivy-lmu: quiet a script :D

why not downloading https://jenkins.ivyteam.io/job/demo-projects/job/master/lastSuccessfulBuild/artifact/connectivity/connectivity-demos/target/connectivity-demos-14.0.0-SNAPSHOT.iar?

I would use adm-zip that we already use for download engine.

playwright/package.json
```diff
@@ -3,12 +3,14 @@
...
     "test:ui": "playwright test"
   },
   "devDependencies": {
     "@playwright/test": "1.61.1",
     "@vscode/test-electron": "3.0.0"
   },
-  "version": "14.0.0"
+  "version": "14.0.0",
+  "dependencies": {
+    "jszip": "^3.10.1"
```
ivy-lmu: adm zip ;)

package.json
ivy-lmu: I would set the env in the jenkins file if possible-> environment {
    CI = 'true'
    RUN_STABLE_VERSION = 'true'
  }

playwright/tests/test.setup.ts
```diff
@@ -0,0 +1,13 @@
+import { test as setup } from '@playwright/test';
+import { runDownloadIar } from './utils/download-iar';
+import { runDownloadAndUnzipVSCode } from './utils/download-vscode';
+
+setup('Setup', async ({}) => {
+  const skipIarDownload = process.env.SKIP_IAR_DOWNLOAD ? true : false;
```
ivy-lmu: do you know which project is triggering the setup?

ivy-lmu (approved):
cool thx 👍

playwright/tests/test.setup.ts
```diff
@@ -0,0 +1,13 @@
...
+  await runDownloadAndUnzipVSCode();
```
ivy-lmu: no need to download vscode if browser integration is running

playwright/tests/utils/download-iar.ts
ivy-dar: I didn't want to hardcode the IAR filename and version, since this will break if we ever release a new version and this is supposed to be an up-to-date iar? I didn't see an easy way to use wildcards in the URL to fetch the "correct" .iar file.

Therefore I went the more complicated route of making the IAR download configurable in the pipeline via `iarDownloadUrl` and `iarFileMustContain` (substrings that the IAR file should contain). To fetch the correct file, I have to download the ZIP and then extract the correct file.

playwright/tests/test.setup.ts
ivy-dar: As far as I understand no, this is not a built in information. The setup project is supposed to run unconditionally. We could have a workaround with env vars, but in that case we're very close to what we're already doing by setting env vars in the `Jenkinsfile` or `package.json` scripts

playwright/tests/page-objects/workspace-page.ts
ivy-dar: Fair point

package.json
ivy-dar: Moved env vars to Jenkinsfiles

playwright/tests/utils/download-iar.ts
ivy-dar: Moved download-iar logic down to the one test that actually uses it. Does fail if iar download fails.

playwright/tests/utils/download-iar.ts
ivy-dar: Removed configurable parameters in pipelines regarding IAR download. Removed download-iar logic from setup, is only used in one test `import-ivy-project.spec.ts`. Moved into test logic and made static.

playwright/package.json
ivy-dar: Removed configurable parameters in pipelines regarding IAR download. Removed download-iar logic from setup, is only used in one test `import-ivy-project.spec.ts`. Moved into test logic and made static.

build/integration/Jenkinsfile
```diff
@@ -22,9 +26,8 @@ pipeline {
             sh """
             pnpm install && pnpm run update:axonivy:next
             pnpm run build:production
-            pnpm run engine:download ${params.engineDownloadUrl}
+            pnpm run engine:download "$ENGINE_DOWNLOAD_URL"
```
ivy-lmu: why do you map to env? ^^

build/integration/browser/Jenkinsfile
ivy-lmu, line 30: here params and env is used

playwright/tests/page-objects/explorer-view.ts
ivy-dar: Yes, all occurences of `hasNode` have been replaced with `hasNodeExact`

build/integration/Jenkinsfile
ivy-dar: good catch, was a leftover

build/integration/browser/Jenkinsfile
ivy-dar: good catch, was a leftover

build/integration/Jenkinsfile
```diff
@@ -11,7 +11,7 @@ pipeline {
   }
 
   parameters {
-    string(name: 'engineDownloadUrl', defaultValue: 'https://jenkins.ivyteam.io/job/core_product/job/master/lastSuccessfulBuild/artifact/workspace/ch.ivyteam.ivy.server.product/target/products/*_Slim_*.zip/*zip*/products.zip', description: 'engine that will be packaged')
+    string(name: 'engineDownloadUrl', defaultValue: 'https://jenkins.ivyteam.io/job/core_product/job/master/lastSuccessfulBuild/artifact/workspace/ch.ivyteam.ivy.server.product/target/products/*_Slim_*.zip/*zip*/products.zip', description: 'engine that will be packaged')
```
ivy-lmu: whitespaces

playwright/tests/fixtures/baseTest.ts
```diff
@@ -103,10 +105,10 @@ const createTmpWorkspace = async (workspace: string) => {
...
   await fs.promises.cp(workspace, tmpWorkspace, { recursive: true });
-  const tmpWsCofig = wsConfig ? path.join(tmpWorkspace, wsConfig) : undefined;
-  return { tmpWorkspace, tmpWsCofig };
+  const tmpWsConfig = wsConfig ? path.join(tmpWorkspace, wsConfig) : undefined;
+  return { tmpWorkspacePath: tmpWorkspace, tmpWsConfig: tmpWsConfig };
 };
 
-const removeTmpWorkspace = async (workspace: string) => {
-  await fs.promises.rm(workspace, { recursive: true, force: true, maxRetries: 3, retryDelay: 1000 });
+const removeTmpWorkspace = async (workspacePath: string) => {
```
ivy-lmu: not sure if it's worth removing the stuff on CI, containers will be removed...

and for windows it might make trobules

ivy-dar: Yeah, we can just not do that

playwright/tests/utils/download-iar.ts
```diff
@@ -0,0 +1,40 @@
...
+  } catch (error) {
+    const systemError = error as NodeJS.ErrnoException;
+    const errorCode = systemError.code ? ` (${systemError.code})` : '';
+    throw new Error(`Failed to write IAR at ${targetPathAbs}${errorCode}: ${systemError.message}`, {
+      cause: error
+    });
+  }
+};
+
+export const runDownloadIar = async (targetPath: string, targetFilename: string, urlIar?: string) => {
```
ivy-lmu: do you call this with urlIar anywhere?

ivy-dar: No, good point I'll simplify this one

playwright/tests/integration/explorer/import-ivy-project.spec.ts
```diff
@@ -0,0 +1,21 @@
...
+import { empty, emptyDownloadIarFilenameUpToDate } from '../../workspaces/workspace';
+
+test.use({ workspace: empty });
+
+test.beforeEach(async ({ tmpWorkspace }) => {
+  await runDownloadIar(tmpWorkspace.tmpWorkspacePath, emptyDownloadIarFilenameUpToDate);
+});
+
+test('Import up-to-date Ivy Project', async ({ wsPage }) => {
+  const explorer = new FileExplorer(wsPage);
```
ivy-lmu, line 16: I think `emptyDownloadIarFilenameUpToDate` is no more global used

ivy-lmu (approved):
cool stuff 😎

ivy-lmu:
btw, when I change test setup, I usually compare all builds where the affected tests are running to check whether the number of executed tests has changed ^^

playwright/tests/integration/explorer/import-ivy-project.spec.ts
ivy-dar: true, removed.

playwright/tests/fixtures/baseTest.ts
```diff
@@ -90,9 +94,6 @@ const runElectronAppTest = async (workspace: string, take: (r: Page) => Promise<
   await page.context().tracing.start({ screenshots: true, snapshots: true, title: test.info().title });
   await take(page);
   await electronApp.close();
-  if (!process.env.CI) {
```
ivy-lmu, line 93: If not CI I would cleanup the temp workspace... but let's do in a followup ;)

extension/src/project-explorer/import-process.ts
ivy-rew: .bpmn and .xml are very common. Some vendors even go for other custom types ... but I don't think we need to support them all. XML would be a nice addition though.

ivy-dar: Perfect, I'll add .xml

PR #1277, opened 2026-07-20
XIVY-19324 make extension deactivation work

ivy-lmu:
### Problem
Extension activation can be slow because it waits for the following call to complete:

```ts
await IvyProjectExplorer.init(context);
```

If VS Code is closed or reloaded before activation finishes, the extension never reaches the activated state. As a result, `deactivate()` is not called, preventing the engine from shutting down.

### Solution

Keep extension activation lightweight by removing the blocking `await` from `activate()`. Instead, defer the required initialization until the project explorer is actually used, where the initialization can be awaited as needed.

### DIFF
--- a/extension/src/project-explorer/ivy-project-tree-data-provider.ts
+++ b/extension/src/project-explorer/ivy-project-tree-data-provider.ts
@@ -99,6 +99,7 @@ export class IvyProjectTreeDataProvider implements TreeDataProvider<Entry> {
         });
       }
     });
+    await IvyProjectExplorer.instance.setProjectExplorerContext({ hasIvyProjects: validProjects.length > 0 });
     return { projects: validProjects.sort(), diagnostics };
   }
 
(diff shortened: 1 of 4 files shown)


extension/src/project-explorer/ivy-project-tree-data-provider.ts
```diff
@@ -152,6 +153,7 @@ export class IvyProjectTreeDataProvider implements TreeDataProvider<Entry> {
   }
 
   async getChildren(element?: Entry): Promise<Entry[]> {
+    await this.activateEnginePromise;
```
ivy-lmu, line 156: vscode tries to load our project explorer, this is one of the first methods called. so lets wait here.

ivy-lli (approved):
makes sense 👍

ivy-dar:
Nice one, thanks!

PR #1284, opened 2026-07-23
XIVY-19233 improve debounce

ivy-lmu:
deploy and invalidate requests are debounced
- we no more need to send a invalidate request, if there is a deploy in the queue (deploy makes invalidate for all projects)
- invalidate is no more possible on project level - all projects will be invalidated


ivy-dar (approved):
LGTM, thanks 👍

ivy-cst (commented):
❤️ 🤞

PR #1296, opened 2026-07-30
try to relaod java projects after conversion

ivy-lmu:


ivy-lli (commented):
When you are currently on this topic, I have some image I would like to see here which I think would improve the UX:

- Show a vscode toast / progess bar which shows how many projects are already converted. Something like `Convert project (5/11)`.
- Stop all watch stuff, so no redeploy or class load invalidate happens while converting

ivy-lmu:
> When you are currently on this topic, I have some image I would like to see here which I think would improve the UX:
> 
>     * Show a vscode toast / progess bar which shows how many projects are already converted. Something like `Convert project (5/11)`.
> 
>     * Stop all watch stuff, so no redeploy or class load invalidate happens while converting

@ivy-lli do you have a sample project that errors due to watch stuff?

ivy-lli:
> @ivy-lli do you have a sample project that errors due to watch stuff?

Portal from v12 to master

ivy-lmu:
I will analyze and do your proposals in a followup

PR #1298, opened 2026-07-31
improve project conversion feedback & make it more stable

ivy-lmu:


extension/src/project-explorer/ivy-project-conversion.ts
```diff
@@ -0,0 +1,72 @@
...
+  let convertedCount = 0;
+  const failedProjects: string[] = [];
+  const numOfProjects = projectsToConvert.length;
+  for (const project of projectsToConvert) {
+    if (token.isCancellationRequested) {
+      logInformationMessage(`Project conversion cancelled by user.`);
+      return;
+    }
+    progress.report({
+      message: `converting ${path.basename(project)} - ${convertedCount} of ${numOfProjects} project(s) converted.${failedProjects.length > 0 ? ` ${failedProjects.length} project(s) failed to convert.` : ''}`
```
ivy-lli:
```suggestion
      message: `Converting ${path.basename(project)} - ${convertedCount} of ${numOfProjects} project(s) converted.${failedProjects.length > 0 ? ` ${failedProjects.length} project(s) failed to convert.` : ''}`
```

extension/src/engine/diagnostics.ts
```diff
@@ -42,11 +42,7 @@ export class IvyDiagnostics {
         if (!fs.existsSync(uri.fsPath)) {
           uri = Uri.joinPath(Uri.file(project.projectDirectory), POM_FILE);
         }
-        let severity = DiagnosticSeverity.Error;
-        if (project.errorMessage?.startsWith(CONVERSION_OUTDATED_MESSAGE_PREFIX)) {
-          severity = DiagnosticSeverity.Warning;
-        }
-        const diagnostic = new Diagnostic(new Range(1, 0, 1, 0), project.errorMessage, severity);
+        const diagnostic = new Diagnostic(new Range(1, 0, 1, 0), project.errorMessage, DiagnosticSeverity.Error);
```
ivy-lmu, line 45: IMO outdated is always an errror

ivy-lli (approved):
Looks great! 🎄

extension/src/engine/diagnostics.ts
ivy-lli: Technically there is a different between `too old` and `outdated`. 
outdated projects should still be able to run without migration
too old needs a convertion.

but I'm not against this change. I think it makes always sense to trigger a convertion, especiall if you are here in the designer

ivy-lmu: I just saw a workspace with a lot of warnings, and it was really not obvious that conversion should be triggered ^^

ivy-lli: yes, true.
I think if this better shows that you should update your project then a very good change :)

PR #1299, opened 2026-07-31
XIVY-19338 Import project - Catch invalid IAR file names

ivy-dar:
- After selecting the IAR file, sanitize and check against all existing Ivy projects in workspace as well as all existing folders on workspace root level.
- Does NOT YET work for .zip, the contents of the .zip are not validated in the frontend. The engine will throw an error, but the frontend will not print a pretty error message.

### DIFF
--- a/extension/src/project-explorer/import-ivy-project.ts
+++ b/extension/src/project-explorer/import-ivy-project.ts
@@ -2,22 +2,27 @@ import path from 'path';
 import { Uri, window, workspace } from 'vscode';
 import type { ImportProjectsBody } from '../engine/api/generated/client';
 import { IvyEngineManager } from '../engine/engine-manager';
+import { listRootsInAllWorkspaces, sanitizeProjectName } from './utils/util';
 
-export const importIvyProject = async (selectedWorkspaceUri: Uri) => {
+export const importIvyProject = async (selectedWorkspaceUri: Uri, allIvyProjectPaths: string[]) => {
   const workspaceId = await IvyEngineManager.instance.getWorkspaceId();
   if (!workspaceId) {
-    window.showErrorMessage('No workspace is currently selected. Please select a workspace before importing an Ivy project.');
     return;
   }
   const selectedTargetPath = selectedWorkspaceUri.fsPath;
-  const input = await collectImportIvyProjectParams();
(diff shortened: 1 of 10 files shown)


extension/src/project-explorer/import-ivy-project.ts
```diff
@@ -1,17 +1,49 @@
...
-  if (input) {
-    await IvyEngineManager.instance.importIvyProject(workspaceId, input);
+export const importIvyProject = async (selectedWorkspaceUri: Uri) => {
+  const activeWorkspaceId = await IvyEngineManager.instance.getWorkspaceId();
+  if (!activeWorkspaceId) {
+    return;
   }
+  const selectedTargetPath = selectedWorkspaceUri.fsPath;
+  const selectedFile = await collectImportIvyArchiveFile();
+  if (!selectedFile || !selectedFile.file) {
```
ivy-lmu: should .filePath be chcked here instead of .file?

extension/src/project-explorer/import-ivy-project.ts
```diff
@@ -1,17 +1,49 @@
...
+  if (existingIvyProjectNames.includes(sanitizedFileName)) {
+    logErrorMessage(
+      `File ${fileImportPath} resolves to project name "${sanitizedFileName}".
+Axon Ivy Project with name "${sanitizedFileName}" already exists in the workspace.
+Please either rename the import file ${fileImportName} or delete/rename the existing project.`
+    );
+    return;
+  }
+
+  const existingRootPaths = (await listRootsInAllWorkspaces()).map(folder => folder.fsPath);
```
ivy-lmu: I would use something like https://www.geeksforgeeks.org/node-js/node-js-fs-existssync-method/

ivy-lmu (approved):
nice 👍

extension/src/project-explorer/import-ivy-project.ts
ivy-dar: I removed the second check altogether, this cannot be undefined in any case I think.

extension/src/project-explorer/import-ivy-project.ts
ivy-dar: my bad, of course that's simpler.  Thanks :+1:

PR #1312, opened 2026-08-07
XIVY-19199 Export ivy project

ivy-lmu:


extension/src/project-explorer/utils/util.ts
```diff
@@ -108,6 +109,13 @@ export const validateNamespace = (value: string) => {
   return 'Enter Namespace separated by "/" -- Only letters, numbers, and underscores are allowed -- Spaces allowed within words -- Empty allowed.';
 };
 
+export const validateExportPath = (fileName: string, folderPath: Uri) => {
+  const filePath = path.join(folderPath.fsPath, `${fileName}.iar`);
+  if (fs.existsSync(filePath)) {
+    return `File already exists: ${filePath}. Please choose a different folder or file name.`;
```
ivy-lli: why can I not override?

ivy-lli (approved):
Why can I not override an existing file?
Idea: move the validate export path method to the execute function. When you press export trigger the export, but first check if the file exists, if yes present a notifcation to the user, the file already exist with abort and override as option

ivy-lmu:
> Why can I not override an existing file? Idea: move the validate export path method to the execute function. When you press export trigger the export, but first check if the file exists, if yes present a notifcation to the user, the file already exist with abort and override as option

let try like this:
image

